PR #3, opened 2017-10-05
Timeout for http.Client and bug fix

gummiboll:
Hello!

This PR adds a default timeout to the http.Client (since a timeout of 0 is == no timeout in go) and adds collector.SetClientTimeout() to override it[1]. Also added a check on err after client.Do since it can cause a panic later on if the request fails.

[1]:
example: `c.SetClientTimeout(5 * time.Second)`

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"time"
 
 	"golang.org/x/net/html"
 
@@ -261,6 +262,11 @@ func (c *Collector) WithTransport(transport *http.Transport) {
 // DisableCookies turns off cookie handling for this collector
 func (c *Collector) DisableCookies() {
 	c.backend.Client.Jar = nil
+}
+
+// SetClientTimeout overrides the default timeout for this collector
+func (c *Collector) SetClientTimeout(timeout time.Duration) {
+	c.backend.Client.Timeout = timeout
 }
 
 func (c *Collector) handleOnRequest(r *Request) {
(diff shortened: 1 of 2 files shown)


colly.go
```diff
@@ -267,6 +268,11 @@ func (c *Collector) DisableCookies() {
 	c.backend.Client.Jar = nil
 }
 
+// SetClientTimeout overrides the default timeout for this collector
```
asciimoo: Could you extend this documentation string with the default timeout value?

colly.go
```diff
@@ -267,6 +268,11 @@ func (c *Collector) DisableCookies() {
...
+func (c *Collector) SetClientTimeout(timeout time.Duration) {
```
asciimoo: What about naming this function simply `SetTimeout` or maybe `SetRequestTimeout`?

http_backend.go
```diff
@@ -106,7 +107,12 @@ func (h *httpBackend) Do(request *http.Request, bodySize int) (*Response, error)
 			<-r.waitChan
 		}(r)
 	}
+
 	res, err := h.Client.Do(request)
+	if err != nil {
```
asciimoo: 👍

colly.go
gummiboll: I actually named it SetTimeout at first but after writing the syntax out it kinda hit me that Collector.SetTimeout() kinda feels like you are setting the timeout on the collector itself and not the outgoing requests. I'll change it to SetRequestTimeout which makes more sense than SetClientTimeout

colly.go
gummiboll: Yes! Good idea!

PR #5, opened 2017-10-06
Fix alignement in example snippet

tomsquest:


README.md
```diff
@@ -24,7 +24,7 @@ With Colly you can easily extract structured data from websites, which can be us
 func main() {
 	c := colly.NewCollector()
 
-    // Find and visit all links
+        // Find and visit all links
```
asciimoo: The problem here is the mix of spaces and tabs, please rewrite it to one tab instead of adding spaces.

tomsquest: Done.

asciimoo:
Thanks.

Issue #9, opened 2017-10-06
Add some basic OnError support


tampajohn:
Probably should break out OnRequestError/OnResponseError, but adding a basic OnError that receives the request, response, and error seems to make sense as a first pass.

Pull request on its way shortly

PR #10, opened 2017-10-06
OnError support, addresses #9

tampajohn:
Adding OnError support to address #9 , also closing res.Body via defer.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -244,6 +244,7 @@ func (c *Collector) OnResponse(f ResponseCallback) {
 	c.lock.Unlock()
 }
 
+// OnError registers a function. Function will be executed on every error.
 func (c *Collector) OnError(f ErrorCallback) {
 	c.lock.Lock()
 	c.errorCallbacks = append(c.errorCallbacks, f)
--- a/colly.go
+++ b/colly.go
@@ -27,6 +27,7 @@ type Collector struct {
 	htmlCallbacks     map[string]HTMLCallback
 	requestCallbacks  []RequestCallback
 	responseCallbacks []ResponseCallback
+	errorCallbacks    []ErrorCallback
 	backend           *httpBackend
 	wg                *sync.WaitGroup
 	lock              *sync.Mutex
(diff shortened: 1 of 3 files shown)


tampajohn:
Well balls, I should have merged down... one sec

examples/bad_request/bad_request.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/asciimoo/colly"
+)
+
+func main() {
+	c := colly.NewCollector()
+
+	c.AllowedDomains = []string{"definitely-note-a-website.not"}
+
+	c.OnHTML("a[href]", func(e *colly.HTMLElement) {
```
asciimoo: Wouldn't be this example more clean and understandable without the `OnHTML` callback?

colly.go
```diff
@@ -213,9 +217,12 @@ func (c *Collector) scrape(u, method string, depth int, requestData map[string]s
 	if len(c.requestCallbacks) > 0 {
 		c.handleOnRequest(request)
 	}
+
 	response, err := c.backend.Do(req, c.MaxBodySize)
-	// TODO add OnError callback to handle these cases
 	if err != nil {
+		if len(c.errorCallbacks) > 0 {
+			c.handleOnError(request, response, err)
```
asciimoo: Perhaps, this should be also called if the response has non-200 status code. What do you think?

examples/bad_request/bad_request.go
tampajohn: Yeah, I just kept this in here to show it wouldn't be called.

colly.go
tampajohn: Makes sense, thoughts on 301/302 redirects?

asciimoo: Redirects are automatically handled by the http client IMO.

https://golang.org/pkg/net/http/#Client

asciimoo: @tampajohn any update on this? I'd like to merge this PR asap because it is a very useful addition, so if you don't have time to code this, i can do it.

tampajohn:
Hey @asciimoo, sorry about the latent response, was out of town/away from tech :)

Thanks for getting the changes in there, looking forward to contributing more later!

PR #15, opened 2017-10-07
Caching when CacheDir is defined

meehow:
I added some basic caching support. Changes should be backward compatible and enable cache only when `CacheDir` is defined. I needed to make `ContextMap` public to make request object serializable with gob.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -31,6 +31,7 @@ type Collector struct {
 	// `0` means unlimited.
 	// The default value for MaxBodySize is 10MB (10 * 1024 * 1024 bytes).
 	MaxBodySize       int
+	CacheDir          string
 	visitedURLs       []string
 	htmlCallbacks     map[string]HTMLCallback
 	requestCallbacks  []RequestCallback
@@ -84,7 +85,7 @@ type HTMLElement struct {
 
 // Context provides a tiny layer for passing data between callbacks
 type Context struct {
-	contextMap map[string]string
+	ContextMap map[string]string
 	lock       *sync.Mutex
 }
 
@@ -107,7 +108,7 @@ func NewCollector() *Collector {
 // NewContext initializes a new Context instance
 func NewContext() *Context {
 	return &Context{
-		contextMap: make(map[string]string),
+		ContextMap: make(map[string]string),
 		lock:       &sync.Mutex{},
 	}
 }
@@ -214,7 +215,7 @@ func (c *Collector) scrape(u, method string, depth int, requestData map[string]s
 	if len(c.requestCallbacks) > 0 {
 		c.handleOnRequest(request)
 	}
-	response, err := c.backend.Do(req, c.MaxBodySize)
+	response, err := c.backend.Cache(req, c.MaxBodySize, c.CacheDir)
 	// TODO add OnError callback to handle these cases
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


colly.go
```diff
@@ -31,6 +31,7 @@ type Collector struct {
 	// `0` means unlimited.
 	// The default value for MaxBodySize is 10MB (10 * 1024 * 1024 bytes).
 	MaxBodySize       int
+	CacheDir          string
```
asciimoo: Please add docstring which describes how `CacheDir` works

colly.go
```diff
@@ -84,7 +85,7 @@ type HTMLElement struct {
 
 // Context provides a tiny layer for passing data between callbacks
 type Context struct {
-	contextMap map[string]string
+	ContextMap map[string]string
```
asciimoo: Why do you make `ContextMap` publicly available? `Context` has a getter/setter which is thread safe.

http_backend.go
```diff
@@ -1,11 +1,16 @@
 package colly
 
 import (
+	"crypto/sha1"
```
asciimoo: sha1 considered unsafe, please use sha256 instead.
(https://www.schneier.com/blog/archives/2017/02/sha-1_collision.html)

colly.go
meehow: I needed to make `ContextMap` public to make request object serializable with gob.

http_backend.go
meehow: sha1 is used here as a convenient way of serializing URL address. Safety is not that crucial here. If somebody can make sha1 collision in purpose, then he probably has many other ways of making web scrapping problematic. Main feature I use from the checksum it's random distribution of first 2 letters (which are used to create a directory).
I made a benchmark (using openssl) and it looks like sha1 is even faster than md5:
```
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes
md5              61097.08k   175900.80k   377240.92k   518029.27k   603403.61k
sha1             74219.17k   208662.02k   485175.72k   726962.18k   846761.59k
sha256           68712.16k   155092.59k   290635.95k   358352.90k   398336.00k
```
I am also aware of possible sha1 collisions, but do you really think that it can be a case here?

asciimoo: You're right, sha1 is totally enough in this case.

colly.go
asciimoo: Cannot be `Reqest.Ctx` excluded somehow?

meehow:
So I added a docstring, we agreed about sha1 and last thing is this `ContextMap`. I know that it's not necessary to make it public, but I don't know how to convince [encoding/gob](https://golang.org/pkg/encoding/gob) to work without having it public.

asciimoo:
> I know that it's not necessary to make it public

The problem here is that it breaks the thread safety of the `Context` object.
Another option is to use https://golang.org/pkg/sync/#Map instead of the default golang `map` which is thread safe by default. It would break the current API, but that's a smaller problem than the thread unsafety IMO.

meehow:
Good point. I will try to solve it.

meehow:
So I solved it by Marshaling and Unmarshaling `Context` to `nil`.

asciimoo:
Perfect && great job, thank you =)

PR #24, opened 2017-10-12
Copy the headers from the last request on redirect

gummiboll:
Apparently headers does not survive a redirect in net/http (https://github.com/golang/go/issues/4800) so here's a fix for that. Probably not the most elegant way of doing it though

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -178,7 +178,6 @@ func (c *Collector) scrape(u, method string, depth int, requestData io.Reader, c
 		return err
 	}
 	req.Header.Set("User-Agent", c.UserAgent)
-
 	if ctx == nil {
 		ctx = NewContext()
 	}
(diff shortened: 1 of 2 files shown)


asciimoo:
@gummiboll great, useful addition. `Authentication` header should be excluded to prevent credential leakage (at least in case the redirect points to a different domain).

gummiboll:
I do get what you mean but as I was thinking about how I would best implement this I started thinking about it; with `AllowedDomains` already in effect, is this needed?

Come to think about it, this PR might override `AllowedDomains`. Will look into it later!

asciimoo:
@gummiboll this redirect handler function could be a private member of the `Collector`, so it could have access to the `AllowedDomains` attribute and it could check if the redirect target is in the `AllowedDomains`. What do you think?

gummiboll:
Good idea! How about this?

asciimoo:
@gummiboll awesome! Could you please exclude `Authentication` header when the redirection points to a different domain?

gummiboll:
Yes, here you go. :)

PR #28, opened 2017-10-13
remove html handler from collector

smileboywtu:


smileboywtu:
example:

``` golang
package main

import (
	"fmt"
	"github.com/asciimoo/colly"
	"strconv"
	"github.com/PuerkitoBio/goquery"
	"time"
	"sync"
	"database/sql"
	_ "github.com/mattn/go-sqlite3"
)

// JianDan duan url template
const JIANDAN_DUAN_URL_TLP = "http://jandan.net/duan/page-%s"

type DBWriter struct {
	lock     *sync.Mutex
	DBfile   string
	instance *sql.DB
}

/**
	Init a new DB writer

	Param filename string: sqlite3 db writer
	Return *DBWriter: dbwriter instance
 */
func NewDBWriter(filename string) *DBWriter {
	db, err := sql.Open("sqlite3", filename)
	if err != nil {
		panic(err)
	}

	return &DBWriter{
		DBfile:   filename,
		lock:     &sync.Mutex{},
		instance: db,
	}
}

/**
	Recreate the table in dababase

 */
func (w *DBWriter) Reinit() {
	sql := `
		DROP TABLE IF EXISTS duanzi;

	    CREATE TABLE duanzi (
        duanzi TEXT NOT NULL,
        created BIGINT NULL);
	`

	_, err := w.instance.Exec(sql)
	w.CheckError(err)
}

/**
	Insert content into database

	Param duanzi string: duanzi
 */
func (w *DBWriter) Write(duanzi string) int64 {
	stmt, err := w.instance.Prepare("INSERT INTO duanzi(duanzi, created) VALUES (?, ?)")
	w.CheckError(err)

	w.lock.Lock()
	res, err := stmt.Exec(duanzi, time.Now().UnixNano())
	w.lock.Unlock()
	w.CheckError(err)

	id, err := res.LastInsertId()
	w.CheckError(err)

	return id
}

/**
	Check if err happens and exit

 */
func (w *DBWriter) CheckError(err error) {
	if err != nil {
		panic(err)
	}
}

/**
	Save data into sqlite 3


 */

/**
	filter for duanzi

	Param a1 string, like vote
	Param a2 string, unlike vote
	Return bool: true if condition is ok
 */
func duan_condition(a1 string, a2 string) bool {

	var a, b int

	if s1, err := strconv.Atoi(a1); err != nil {
		return false
	} else {
		a = s1
	}

	if s2, err := strconv.Atoi(a2); err != nil {
		return false
	} else {
		a = s2
	}
	return a > 2*b
}

/**
	find the max tab size of duan at jiandan.com

	Param selector *goquery.Selection: Go Query Selection, used as Jquery like way
	Return int: largest page size
 */
func find_largest_page(selector *goquery.Selection) int {
	page_str := selector.Find("span").First().Text()
	if page, err := strconv.Atoi(page_str[1:len(page_str)-1]); err == nil {
		return page
	}
	return 0
}

/**
	Generate Url template

	Param page string: current page size
	Return url string: visit url
 */
func generate_url(page string) string {
	return fmt.Sprintf(JIANDAN_DUAN_URL_TLP, page)
}

func main() {
	c := colly.NewCollector()
	c.AllowedDomains = []string{"jandan.net"}
	c.SetRequestTimeout(5 * time.Second)
	c.Limit(&colly.LimitRule{
		DomainGlob:  "*jiandan.net*",
		Parallelism: 5,
	})

	w := NewDBWriter("test.db")
	w.Reinit()

	var page_size int

	// find the page number
	c.OnHTML(".cp-pagenavi", func(e *colly.HTMLElement) {
		page_size = find_largest_page(e.DOM)
	})
	c.Visit("http://jandan.net/duan/")
	c.OnHTMLDetach(".cp-pagenavi")

	// scrape duanzi
	c.OnHTML(".row", func(e *colly.HTMLElement) {
		like_str := e.DOM.Find(".tucao-like-container>span").Text()
		unlike_str := e.DOM.Find(".tucao-unlike-container>span").Text()
		if duan_condition(like_str, unlike_str) {
			duanzi := e.DOM.Find(".text>p").Text()
			w.Write(duanzi)
		}
	})

	for i := page_size; i > 0; i-- {
		go c.Visit(generate_url(strconv.Itoa(i)))
	}

	c.Wait()
}
```

asciimoo:
@smileboywtu thank you. This addition can be handy in some situations, however you have another option in the above example: use two collectors, one for extracting the page number and one for scraping the rest.

colly.go
```diff
@@ -269,6 +269,13 @@ func (c *Collector) OnHTML(goquerySelector string, f HTMLCallback) {
 	c.lock.Unlock()
 }
 
+// OnHTML deregister a function. Function will not be execute after register
```
asciimoo: Please adjust the function name in comment, it is `OnHTMLDetach` and not `OnHTML`

hondajojo:
@asciimoo if use two collector, how to pass ctx  or attribute like Scrapy's meta between collectors

asciimoo:
> @asciimoo if use two collector, how to pass ctx or attribute like Scrapy's meta between collectors

You're right, it has limitations, that's why I wrote that your addition can be useful sometimes.
I'd like to merge the PR, but please fix the function name in docstring before.

smileboywtu:
please review it again

Issue #29, opened 2017-10-14
I think  c.Limit  also can  be used to filter unmatched urls  . Should support regex  before request   in anywhere. 


tx991020:
The purpose is to filter out the mismatched urls

for example
package main

import (
	"github.com/asciimoo/colly"
	"fmt"
	"time"
)

func main() {
	urls := []string{"https://httpbin.org/hello", "https://httpbin.org/123", "https://httpbin.org/12xyz"}

	// Instantiate default collector
	c := colly.NewCollector()

	// when visiting links which domains' matches "*httpbin.*" glob
	c.Limit(&colly.LimitRule{
		DomainGlob: "*https://httpbin.org/[a-z]+",
	})
	c.OnResponse(func(r *colly.Response) {
		fmt.Println("Finished", r.Request.URL, time.Now())
	})

	for _, i := range urls {
		c.Visit(i)
	}
	

	
     c.Wait()
}
Starting https://httpbin.org/hello",
Starting "https://httpbin.org/123",
Starting "https://httpbin.org/12xyz
Finished  https://httpbin.org/hello

asciimoo:
@tx991020 this would be a useful feature, however `LimitRule` works a bit different than in your example. Currently it specifies **restrictions** to the matching domains and you want to **allow** URLs with these rules. I'm not sure if extending `LimitRule` is the correct way to achieve this functionality.

asciimoo:
@tx991020 I could imagine this as an independent feature - e.g. `c.AddURLFilter(urlGlob string)` or something similar. Thoughts?

tx991020:
yes.

meehow:
In my understanding `Limit` it's something what limits resources used by a program. You need to extract URLs in code of your program anyway.
Isn't your code a good place to add some regexp?

meehow:
I would rather add `c.URLFilters []*regexp.Regexp` to make sure before crawling that regexps are correct.

PR #31, opened 2017-10-15
handling of DisallowedDomains

meehow:
Hi,
my crawled stated to crawl some social media pages, which are linked on almost every page, so I thought that it would be nice if I can add list of domains which I don't want to crawl.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -29,6 +29,8 @@ type Collector struct {
 	// AllowedDomains is a domain whitelist.
 	// Leave it blank to allow any domains to be visited
 	AllowedDomains []string
+	// DisallowedDomains is a domain blacklist.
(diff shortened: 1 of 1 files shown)


asciimoo:
@meehow great, this feature makes sense, however it would be good to generalize somehow the domain/URL filtering to be able to implement the functionality requested in #29 too. Thoughts?

meehow:
I'm not sure about #29. `DisallowedDomains` was easy to add because `AllowedDomains` already exists. Maybe these 2 should go to separate URL filtering feature. Plus of this solution would be that it can be integrated with `checkRedirectFunc`.

asciimoo:
@meehow agreed. Let's implement URL filtering as an independent feature. Thanks!

Issue #8, opened 2017-10-06
Multipart support


kawakami-o3:
Do you have any plan to support a multipart POST request?
I will code this feature later.

asciimoo:
@kawakami-o3 this is definitely a missing and useful feature. I'd appreciate if you could implement it.

vosmith:
So I started working on this and gave it a good try.  The main issue that there is no way to set custom headers.

Right now all `POST` requests are forced into the content-type `application/x-www-form-urlencoded`.

I tried creating new functions for like 

    func (c *Collector) PostMultipart ...

But once it hits `scrape`, the content gets set.  I also tried adding an `onRequest` callback method to set the header, but then this callback affects all successive requests once it becomes a part of the callback stack.

Frankly I found it difficult to accomplish this without significant changes to how the headers are handled.

asciimoo:
> So I started working on this and gave it a good try. 

Awesome! =)

> Right now all POST requests are forced into the content-type application/x-www-form-urlencoded

What about adding a new "Content-Type" string parameter to `scrape()` function and in case it is not empty we set the content type to that?

vosmith:
Sounds like it should work! I'll go ahead and give it a try👍🏾

kawakami-o3:
I am late...
I planed to add a request data handler instead of `requestData map[string]string`.
If the handler contains a file, it can decide correct Content-Type.

Should I stop my work?

asciimoo:
@kawakami-o3 seems like @vosmith nearly finished the implementation. Let's wait for his solution to avoid duplicated work. What do you think?

kawakami-o3:
I follow you.

asciimoo:
added in 30e15f6 thank you for your work.

PR #33, opened 2017-10-16
initial work on multipart requests, issue #8

vosmith:
Here is the initial work for performing multipart post requests. added to the API are:

    func (c *Collector) PostMultipartRaw(URL string, requestData map[string][]byte) error
    func (r *Request) PostMultipartRaw(URL string, requestData map[string][]byte) error

Internally, `scrape` was modified to support custom headers and a helper function for fomatting the multipart request body was also added.

An example can be found in `examples/multipart.go`


asciimoo:
Good PR, thanks!
Wouldn't be simply `PostMultipart` a better name instead of `PostMultipartRaw`? What is the meaning of "raw" here?

vosmith:
I used `raw` to follow the precedent set by the `PostRaw` method where the payload is a `[]byte` as opposed to a string.  I'm not too attached to it though.

asciimoo:
`PostRaw` is `raw` because it receives "raw" data and not a structured one. `PostMultipart(Raw)`'s parameter is a key-value data structure which is not "raw" to me. ..And `PostMultipart` is shorter =)

vosmith:
No problem, will make that change right now

PR #34, opened 2017-10-17
Add Collector.Post test

kawakami-o3:


kawakami-o3:
Thank you for your advice.
I removed the `.gitignore` file.

Issue #37, opened 2017-10-20
Proxy Support


vosmith:
What are your thoughts on adding proxy support to the `Collector`?  I see that one could just create a custom `Transport` and set the collector to use it, but it would be nice to have a `SetProxy(url)` method or something similar.

vosmith:
I think this issue is closed by c1a71b6cb01c12f2d1ff61425d06b5ae67b5f512

asciimoo:
Yup, thanks @gummiboll

PR #38, opened 2017-10-21
Adds SetProxy (regarding #37)

gummiboll:


colly.go
```diff
@@ -346,6 +346,20 @@ func (c *Collector) SetRequestTimeout(timeout time.Duration) {
...
 }
 
+// SetProxy sets a proxy for the collector
+func (c *Collector) SetProxy(proxyURL string) error {
+	proxyParsed, err := url.Parse(proxyURL)
+	if err != nil {
+		return err
+	}
+
+	c.backend.Client.Transport = &http.Transport{
```
asciimoo: Why do you initialize a new `Transport` object for this?
Wouldn't `c.backend.Client.Transport.Proxy = http.ProxyURL(proxyParsed)` be enough?

gummiboll: Because apparently, you can't: https://play.golang.org/p/5jR16wr0Ca - if I understand it correctly a http.Client gets a RoundTripper-transport by default if none is specified that doesnt have a Proxy-field

asciimoo: Ah, I didn't know that `http.Client.Transport` is an interface. Then, we could check if `c.backend.Client.Transport` interface is a default `*http.Transport` type. If it is, we can set the proxy without overwriting the whole `Transport`. Something like this:
```go
	t, ok := c.backend.Client.Transport.(*http.Transport)
	if client.Transport != nil && ok {
		t.Proxy = http.ProxyURL(proxyURL)
	} else {
		<<create new transport>>
	}
```
The purpose of my suggestion is to avoid overriding the previously set `Transport` if possible.
Also, it would be good to document this behavior in the docstring, because the previously set custom `http.RoundTripper` will be overwritten even with the suggested mods.

gummiboll: How about this?

PR #62, opened 2017-12-11
Implement OnScraped callback

llonchj:
#61

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -491,8 +491,8 @@ func (c *Collector) OnError(f ErrorCallback) {
 	c.lock.Unlock()
 }
 
-// OnScraped registers a function. Function will be executed at the end
-// of a scraping.
+// OnScraped registers a function. Function will be executed after
+// OnHTML, as a final part of the scraping.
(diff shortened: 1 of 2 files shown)


asciimoo:
What is the purpose of `OnScraped` callback? Seems like it is identical to `OnResponse` callback.

llonchj:
Because I collect information in multiple `OnHTML` callbacks and store data in Ctx, the plan is to use `OnScraped` to perform a final processing of Ctx (aka Scraped item pipeline)

asciimoo:
Ah, I see, so the difference between `OnScraped` and `OnResponse` is that `OnScraped` is called after the `OnHTML` callbacks. Could you please describe it in the docstring of `OnScraped`?

llonchj:
@asciimoo done!

llonchj:
most welcome @asciimoo. Thanks for colly, it's an amazing tool.

ash-kamrip:
There should be separate example for this , explainging the usage of multiple `OnHTML` callbacks and then using the  `OnScraped`

PR #63, opened 2017-12-14
NewHTMLElementFromSelectionNode implementation

llonchj:
Case scenario: Using HTMLElement in external package unittest

Problem: HTMLElement has local `attributes` that make impossible from other packages to instantiate a HTMLElement with values in attributes.

FIx: Implementation of NewHTMLElementFromSelectionNode


llonchj:
Thanks @asciimoo!

PR #66, opened 2017-12-18
Become a contributor link broken in README.md

ElijahLynn:


asciimoo:
Thanks for your effort but @SeanSanker submitted the same changes 4 minutes before you (#65), so I pulled that.

Issue #54, opened 2017-11-26
Randomized Delay


jodizzle:
Is it possible to create randomized delays, i.e., per-request delays selected from some range or based on some random factor?  I couldn't think of a good way to do this, other than maybe cycling through several `Collector`s with different limit sets, which seems sub-optimal.

It seems like having `LimitRule.DelayRange` or `LimitRule.RandomFactor` options would be quite helpful.

asciimoo:
@jodizzle this would be a handy addition. Do you want to work on it?

As a workaround you can add an OnRequest callback with random sleeps:

```go
c.OnRequest(func(_ *colly.Request) {
    time.Sleep(time.Duration(rand.Randint(5)+1) * time.Second)
})
```

jodizzle:
@asciimoo Ah, I hadn't thought of the `OnRequest` approach.  Thanks.

And sure, I can try working on it.  Do you have any thoughts about what the best implementation would be?

asciimoo:
I like the idea of extending `LimitRule` with a `RandomDelay time.Duration` attribute which is 0 by default. If its value isn't 0, a random duration between 0 and `RandomDelay` could be added to `LimitRule.Delay` on every request. This solution doesn't break the current API and it's intuitive for me. What do you think?

jodizzle:
That sounds pretty reasonable.  I'll take a look.

jlr52:
Not sure if its okay, but I got interested in the code base so I just implemented the enhancement.

If @jodizzle still want to fix it I can decline my pull request.

Let me know.

asciimoo:
implemented in #67 - thanks @jlr52

PR #67, opened 2017-12-19
Added random delay per issue #54

jlr52:
Added random delay config to limitRule that allow user to specify additional parameter to
        randomized the delay on each request.

Testing Methodogy:

    Used the rate_limit app in the _example dir and verify that adding RandomDelay will cause
        each request to have different delay time.

### DIFF
--- a/http_backend.go
+++ b/http_backend.go
@@ -78,6 +78,7 @@ func (r *LimitRule) Init() error {
 }
 
 func (h *httpBackend) Init() {
+	rand.Seed(time.Now().UnixNano())
 	h.LimitRules = make([]*LimitRule, 0, 8)
 	jar, _ := cookiejar.New(nil)
 	h.Client = &http.Client{
@@ -152,7 +153,6 @@ func (h *httpBackend) Do(request *http.Request, bodySize int) (*Response, error)
 		r.waitChan <- true
 		defer func(r *LimitRule) {
 			randomDelay := time.Duration(0)
-			rand.Seed(time.Now().UnixNano())
 			if r.RandomDelay != 0 {
 				randomDelay = time.Duration(rand.Intn(int(r.RandomDelay)))
 			}
--- a/http_backend.go
+++ b/http_backend.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"net/http/cookiejar"
 	"os"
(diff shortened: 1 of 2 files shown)


asciimoo:
Please fix go formatting issues by running `go fmt`. (https://travis-ci.org/gocolly/colly/builds/318512944)

jodizzle:
I don't think the conditional is necessary.  It seems like you could just do
```go
randomDelay := time.Duration(rand.Intn(int(r.RandomDelay)))
```
right?

Also, I think it would be nice to either expand the `rate_limit.go` example or add another example in that same folder covering use of `RandomDelay`.

jlr52:
@asciimoo  good call on go fmt , the file is now formatted

@jodizzle good call, I added another example. And the reason I need the condition is because

    `invalid argument to Intn` will happen if `0` is the input of `rand`

http_backend.go
```diff
@@ -148,7 +151,12 @@ func (h *httpBackend) Do(request *http.Request, bodySize int) (*Response, error)
 	if r != nil {
 		r.waitChan <- true
 		defer func(r *LimitRule) {
-			time.Sleep(r.Delay)
+			randomDelay := time.Duration(0)
+			rand.Seed(time.Now().UnixNano())
```
asciimoo: Perhaps this line should be removed to be able to control the seed externally. It can be useful for testing and for reproducible runs. What do you think?

jlr52: I think that is a good idea, however seed can't be static, so I think we need to add something like this

` type LimitRule struct { SeedFunc func() int64 }`

what do you think?

asciimoo: > I think that is a good idea, however seed can't be static

You're right.
What about calling `rand.Seed(time.Now().UnixNano())` from the package's `init()` function which runs only on startup?

jlr52: hmm, the seed has to be regenerate per request to achieve per-request random delays, generating a static seed in `init` might not work, or maybe I am misunderstanding?

We could also make all request having the same delay and make the randomness per process.

What do you think?

asciimoo: Seed is used to initialize random generation. This little example demonstrates how it works: https://play.golang.com/p/4xks-DcCgv

More info: https://en.wikipedia.org/wiki/Random_seed

jlr52: yeah, you are right, for some reason I was thinking that I need to re seed to make each request random. Let me put the seed in init

PR #69, opened 2017-12-20
added items pipeline support

nasirhjafri:
Added support to add pipelines to that scraped items can be process on the go. 

i.e. can be used for data cleaning, adding items to the production for continuous integration, collecting stats or printing items while we are crawling.

Added example for pipeline in `_examples/pipelines/pipelines.go`

### DIFF
--- a/_examples/pipelines/pipelines.go
+++ b/_examples/pipelines/pipelines.go
@@ -1,0 +1,118 @@
+package main
+
+import (
+	"log"
+	"strings"
+
+	"github.com/PuerkitoBio/goquery"
+	"../../../colly"
+	"encoding/json"
+)
+
+// Course stores information about a coursera course
+type Course struct {
+	Title       string
+	Description string
+	Creator     string
+	Level       string
+	URL         string
+	Language    string
+	Commitment  string
+	HowToPass   string
+	Rating      string
+}
+
+var total_items int = 0
+
+func main() {
+	// Instantiate default collector
+	c := colly.NewCollector()
+	//Example stats pipeline
+	c.AddPipeline(func(item interface{}) interface{} {
+		total_items++
+		log.Println("Total item scrapped: ", total_items)
+		return item
+	})
+
+	//Example json serializer pipeline
+	c.AddPipeline(func(item interface{}) interface{} {
+		jsonData, _ := json.MarshalIndent(item, "", "  ")
+		log.Println("Item scraped: ", item)
+		return jsonData
+	})
+
+	// Visit only domains: coursera.org, www.coursera.org
+	c.AllowedDomains = []string{"coursera.org", "www.coursera.org"}
+
+	// Cache responses to prevent multiple download of pages
+	// even if the collector is restarted
+	c.CacheDir = "./coursera_cache"
+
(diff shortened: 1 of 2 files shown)


nasirhjafri:
hi @asciimoo should i remove the example, as example is using the latest code and due to relative path to import the latest Collector class the tests are failing?

asciimoo:
hello @nasirhjafri thank you for your contribution.
What is the benefit of having this pipeline inside colly? As I see it doesn't use colly's internals so it can be easily implemented in the scrapers if needed.

nasirhjafri:
Hello @asciimoo As `colly` is a scraping library it should provide all scraping libraries features like pipelines, custom middle wares, support of xpath, in which currently `colly` lacks likewise other scraping libraries i.e. scrapy for python has all of these. To get `colly` used by more people and companies we need to provide such features so that they could find its easy to use instead of writing every thing by them selves. 

some typical uses of item pipelines are:

cleansing HTML data
validating scraped data (checking that the items contain certain fields)
checking for duplicates (and dropping them)
storing the scraped item in a database

nasirhjafri:
Moreover pipelines can help to separate data cleaning/ processing code from data extraction. In future we need to add support for pipelines to run in background processes  after an item is scrapped and passed to pipelines. As no one wants right now to write code for pipelines and code to run it in background separate from crawling process. If the user of library has to do so much effort they might find some other alternatives meanwhile.

asciimoo:
A benefit of the callback based design is that you can create your own pipelines out of the box. If you move the body of the pipeline functions in the example to the `OnHTML` callback where you call `c.ItemScraped()` it provides the same functionality and perhaps it requires less boilerplate. Another benefit of handling these pipelines in the scraper is that we don't need to do type conversions from `interface{}`. Furthermore you can define and call these pipeline functions from the callbacks without adding it to colly.

Example:

```go
func serialize(item *Course) {
	jsonData, _ := json.MarshalIndent(item, "", "  ")
	log.Println("Item scraped: ", item)
})

func stats(item *Course) {
	total_items++
	log.Println("Total item scrapped: ", total_items)
})

func pipeline(c *Course) {
	stats(c)
	serialize(c)
}

detailCollector.OnHTML(`div[id=rendered-content]`, func(e *colly.HTMLElement) {
	// [...]
	pipeline(course)
	// you can even run it in background with "go pipeline(course)"
}
```

> Moreover pipelines can help to separate data cleaning/ processing code from data extraction

This can be achieved by defining a function in your scraper, see the above example.

> As colly is a scraping library it should provide all scraping libraries features like pipelines, custom middle wares, support of xpath

I agree on xpath and partially on middlewares, but pipelines can be created without this addition IMO.

I still can't see the benefit of this addition. It doesn't add extra functionality nor simplify the usage of colly. Feel free to reopen if you disagree.

LucaPaterlini:
Sorry for reopening on this topic but can it be possible to have a middleware, in my case I would like to be able to choose the function I prefer to download the page.
eg. in this case I would like to be able to add this middleware so I can have the js geerated pages rendered,
https://github.com/geziyor/geziyor#making-js-rendered-requests

PR #71, opened 2017-12-21
[fix] close file after checking error in http backend

ferhatelmas:
If create returns error, file will be nil so panics. Defer moved after error check.


asciimoo:
What does this pr fix?

ferhatelmas:
@asciimoo updated PR description.

PR #78, opened 2018-01-03
Add iteration for context

rongyi:


context_test.go
```diff
@@ -0,0 +1,23 @@
...
+	for i := 0; i < 10; i++ {
+		ctx.Put(strconv.Itoa(i), i)
+	}
+	values := ctx.ForEach(func(k string, v interface{}) interface{} {
+		return fmt.Sprintf("%s ==> %d", k, v.(int))
+	})
+	if len(values) != 10 {
+		t.Fatal("fail to iterate context")
+	}
+	for _, s := range values {
```
asciimoo: This logging should be removed imo. It's only useful for debug purposes, doesn't add value to the test.

rongyi: Sorry, I forgot to format the code. I will send you a new PR

PR #84, opened 2018-01-05
Examples with functional options

peterhellberg:
Use functional options for NewCollector in examples

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ go:
   - 1.8.x
   - 1.9.x
   - tip
-env:
-  - "PATH=$GOPATH/bin:$PATH"
 script:
   - go get -u github.com/golang/lint/golint
   - OUT="$(go get -a)"; test -z "$OUT" || (echo "$OUT" && return 1)
--- a/_examples/basic/basic.go
+++ b/_examples/basic/basic.go
@@ -8,10 +8,10 @@ import (
 
 func main() {
 	// Instantiate default collector
-	c := colly.NewCollector()
-
-	// Visit only domains: hackerspaces.org, wiki.hackerspaces.org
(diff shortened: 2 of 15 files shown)


peterhellberg:
Travis-CI build failed due to connection problems between Travis-CI and GitHub when starting the `tip` build.

.travis.yml
```diff
@@ -1,12 +1,18 @@
...
+
 go:
   - 1.7.x
   - 1.8.x
   - 1.9.x
   - tip
+
+sudo: false
+
+dist: trusty
```
asciimoo: What's the benefit of explicitly declaring trusty? We have to manually update this if trusty becomes obsolete.

peterhellberg: I thought it would be needed in order to force the use of the Container-based build environment, but maybe not.

asciimoo: When I added `sudo: false` I noticed speedup immediately and if I compare the duration of this PR's travis job and the one where `sudo: false` was intrudoced I cannot see significant differences:
 - https://travis-ci.org/gocolly/colly/builds/325491037
 - https://travis-ci.org/gocolly/colly/builds/325611501

peterhellberg: Then maybe I should remove that commit? (I also wonder why `"PATH=$GOPATH/bin:$PATH"` is set in the env section, `$GOPATH/bin` is automatically added to the $PATH by Travis-CI)

asciimoo: You're right, then could you remove these lines?
If it's done, I'm merging this PR and migrating the new examples to the documentation site (I have a "beautiful" oneliner for this purpose =] ).

peterhellberg:
I've replaced the commit changing `.travis.yml` with one that removes the env setting `"PATH=$GOPATH/bin:$PATH"`

PR #89, opened 2018-01-13
WIP: xpath support

vosmith:
In #69 it was mentioned that xpath support would be welcomed.  I've started some work on this to see if this is what you had in mind.  Let me know what you think!


asciimoo:
@vosmith this is super awesome, thanks. I have not heared about `github.com/antchfx/xmlquery` before, but looks promising.
As I see, it isn't compatible with html, however @antchfx has a dedicated package for querying html: github.com/antchfx/htmlquery .

vosmith:
Yeah no problem, I think the two libraries have a very similar API.  I'll switch it over and make the necessary adjustments.

asciimoo:
@vosmith great, thank you! Tests and benchmarks would be appreciated. =]

vosmith:
I converted from the xmlquery library to the htmlquery library.  Let me know what you think!

asciimoo:
@vosmith thanks for the changes. Something went wrong with the rebase and the PR contains previous commits, please fix it. Also, what about using `xmlquery` for XML content type and `htmlquery` for HTML content inside `handleOnXML`?

vosmith:
Yeah i rebased to the master branch and it pulled in all of those commits, I'll try to get rid of them to clean this up.

Also I do like the idea of using different parsers based on the content-type.  I might be able to do some example crawling sitemaps as well.  Like Shopify https://www.shopify.com/sitemap.xml.  Let me see what I can do.

vosmith:
I've updated the `handleOnXML` to support `xmlquery` for `text/xml` content types and `htmlquery` for `text/html` content types.  I've also added the Shopify sitemap example in the` _examples/sitemap` directory.  Let me know what you think!

colly.go
```diff
@@ -688,6 +732,50 @@ func (c *Collector) handleOnHTML(resp *Response) {
 	}
 }
 
+func (c *Collector) handleOnXML(resp *Response) {
+	if (!strings.Contains(strings.ToLower(resp.Headers.Get("Content-Type")), "html") && !strings.Contains(strings.ToLower(resp.Headers.Get("Content-Type")), "xml")) || len(c.xmlCallbacks) == 0 {
+		return
+	}
+
+	if !strings.Contains(strings.ToLower(resp.Headers.Get("Content-Type")), "html") {
```
asciimoo: This condition is a bit unclear for me. Why do you negate the expression? As I understand this block is executed if the content-type doesn't contain "html".

colly.go
```diff
@@ -688,6 +732,50 @@ func (c *Collector) handleOnHTML(resp *Response) {
...
+				if c.debugger != nil {
+					c.debugger.Event(createEvent("xml", resp.Request.ID, c.ID, map[string]string{
+						"selector": cc.Query,
+						"url":      resp.Request.URL.String(),
+					}))
+				}
+				cc.Function(e)
+			})
+		}
+	} else if !strings.Contains(strings.ToLower(resp.Headers.Get("Content-Type")), "xml") {
```
asciimoo: I have the same question here as in the previous block.

_examples/sitemap/sitemap.go
```diff
@@ -0,0 +1,29 @@
+package main
```
asciimoo: Perhaps this example could have a bit specific name, e.g. `shopify_sitemap`. `sitemap` is too general for me. What do you think?

colly.go
vosmith: Oh my! Good catch.

PR #100, opened 2018-02-02
Twitter user profile

bharatsewani1993:
This is example script to fetch twitter users full profile details including, Full Name, Last Name, Username, Join Date, Followers, Following, Likes, Tweets, Mentions, Lists, Location, Website, Bio, If an account is verified.
Based on that data users could run a deep analysis of profile type and other details like gender and profile weight and impact prediction.

### DIFF
--- a/_examples/Twitter_userprofile/main.go
+++ b/_examples/Twitter_userprofile/main.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"github.com/gocolly/colly"
+)
+
+
+// Userprofile store twitter user data
+type Userprofile struct {
+	FullName string
+	UserName string
+	AccountVerified string
+	Bio string
+	Location string
+	Website string
+	JoinDate string
+	Tweets string
+	Following string
+	Followers string
+	Likes string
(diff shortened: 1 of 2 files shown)


asciimoo:
@bharatsewani1993 Thanks for the new example. Seems there are code formatting issues: https://travis-ci.org/gocolly/colly/builds/336403265 , please fix those.

Also, I have some ideas to improve the scraper:
 - The twitter usernames could be command line/stdin parameters
 - Use integer types for `followers`, `following`, `likes`, `tweets`, `lists` fields
 - Use less fragile css selectors. (e.g. `"div#page-container > div:nth-of-type(2) > div > div > div > div > div > div > div > h1 > span > a > span > span"` breaks if any of the listed elements changes)
 - Write the json result to the standard output, so it can be piped to other commands or redirected to a file.

asciimoo:
@bharatsewani1993 as I see, you already created a repo for this scraper: https://github.com/bharatsewani1993/Twitter_user_profile_scrapper . Please refer to your project in our README.md instead of adding it as an example. This example doesn't contain anything new compared to the other examples and we have a dedicated section for projects using colly (https://github.com/gocolly/colly#other-projects-using-colly).

PR #102, opened 2018-02-07
Introduces reddit example

frankcash:
This pull request introduces an example for calling reddit.com/r/programming

This code has some interesting features.  

- It does some random time delaying
- Limits the parallelism
- Creates a generic struct to store crawl information
- Returns a mass splice with all stored crawl information at end of crawl
- Hits up the next page element to crawl the next page (if there are more) for more information!

I would appreciate if this example was introduced into the code-base!

If there are any critiques or changes needed to get it in please let me know!

### DIFF
--- a/_examples/reddit/reddit.go
+++ b/_examples/reddit/reddit.go
@@ -25,7 +25,7 @@ func main() {
 	)
 
 	// On every a element which has .top-matter attribute call callback
-  // This class is unique to the div that holds all information about a story
+	// This class is unique to the div that holds all information about a story
 	c.OnHTML(".top-matter", func(e *colly.HTMLElement) {
 		temp := item{}
 		temp.StoryURL = e.ChildAttr("a[data-event-action=title]", "href")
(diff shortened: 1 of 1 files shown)


_examples/reddit/reddit.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// On every span tag with the class next-button
+	c.OnHTML("span.next-button", func(h *colly.HTMLElement) {
+		t := h.ChildAttr("a", "href")
+		c.Visit(t)
+	})
+
+	// Set max Parallelism and introduce a Random Delay
+	c.Limit(&colly.LimitRule{
+		Parallelism: 2,
```
asciimoo: Parallelism has no effect if the application is single threaded. Use goroutines or `c.Async = true` to achieve parallel requests.

frankcash:
LIT

Just so it is known if people cannot figure it out to run

```
$ go run reddit.go <subreddit> <subreddit> <subreddit> <etc>
```

Issue #114, opened 2018-03-16
Redirect fine grained control


llonchj:
I need to change my logic based on the status response of a particular URL: 200 or 301.
How can I get control of redirects in a collector?

PR #115, opened 2018-03-16
collector.FollowRedirects support

llonchj:
Support for #114

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -57,6 +57,9 @@ type Collector struct {
 	// MaxDepth limits the recursion depth of visited URLs.
 	// Set it to 0 for infinite recursion (default).
 	MaxDepth int
+	// FollowRedirects allows Visit to handle redirects automatically
+	// Set it to false for the collector to handle 30x responses.
+	FollowRedirects bool
 	// AllowedDomains is a domain whitelist.
 	// Leave it blank to allow any domains to be visited
 	AllowedDomains []string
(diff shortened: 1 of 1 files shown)


asciimoo:
@llonchj Thanks for the PR. Wouldn't be better to add a `Collector.RedirectHandler func(req *http.Request, via []*http.Request) error` member instead of `Collector.FollowRedirects bool`? It would be more flexible, but requires a bit more boilerplate.
Your use case could be implemented like this:
```go
c := colly.NewCollector()
c.RedirectHandler = func(req *http.Request, via []*http.Request) error {
    return http.ErrUseLastResponse
}
```
or a simple logging:
```go
c.RedirectHandler = func(req *http.Request, via []*http.Request) error {
    log.Println("redirect", req.URL, via[len(via)-1].URL)
}
```
What do you think?

llonchj:
I did not want to interfere with the current checkrrdirect func.

And the handler can’t be set uninflected a environment variable.

I can go whichever way. Just think the redirect has to be configurable.

What do you want me to do?

asciimoo:
> I did not want to interfere with the current checkrrdirect func.

That's a good approach, it could be the same with the `RedirectHandler`. `checkRedirectFunc` could return the `RedirectHandler` if it isn't `nil` or the current function otherwise.

Perhaps it would be even better to call `RedirectHandler` in the beginning of the current redirect checker function and return immediately if the `RedirectHandler` returns an error.

> And the handler can’t be set uninflected a environment variable.

The environment could be handled similarly: 

```go
"FOLLOW_REDIRECTS": func(c *Collector, val string) {
	if !isYesString(val) {
		c.RedirectHandler = func(req *http.Request, via []*http.Request) error {
			return http.ErrUseLastResponse
		}
	}
},
```

> What do you want me to do?

Generally I prefer to choose the more flexible solutions, but I'm not sure now.. Both solutions have pros and cons. What about trying the `RedirectHandler` approach and iterate if it isn't intuitive enough?

llonchj:
Makes sense, will do. Thanks for your vision

2018-03-17 6:06 GMT+11:00 Adam Tauber <notifications@github.com>:

llonchj:
@asciimoo, changes pushed following your advise. it looks good. what do you think?

llonchj:
Thanks for your support, @asciimoo! It has been a pleasure

2018-03-17 6:31 GMT+11:00 Adam Tauber <notifications@github.com>:

PR #118, opened 2018-03-21
add missing collector elements in collector.Clone

llonchj:


llonchj:
My pleasure @asciimoo!

PR #134, opened 2018-05-08
Fix request.Retry() ignoring AllowURLRevisit

gsoec:


asciimoo:
This is a bug indeed, but what about using `r.collector.AllowURLRevisit` or a constant `true` instead of defining a parameter for `r.Retry()`? `false` actually makes no sense, because the URL is already visited, that's why we want to call `Retry()`.

gsoec:
```go
func main() {
  c := colly.NewCollector()

  retryCount := 0

  c.OnError(func(r *colly.Response, err error) {
    if retryCount < 3 {
      retryCount += 1

      // c.AllowURLRevisit = true
      // r.Retry()

      // ignore AllowURLRevisit and force fetch for this request only
      r.Retry(false)
    }
  })

  c.Visit("https://definitely-not-a.website/")
}
```
I am retrying specific request with setting `c.AllowURLRevisit = true`
and not sure if it's influencing the other requests in the queue.
`r.Retry(false)` might work in this case ? i guess

asciimoo:
Correct me if I'm wrong, but I think we can simply change `return r.collector.scrape(r.URL.String(), r.Method, r.Depth, r.Body, r.Ctx, *r.Headers, false)` to `return r.collector.scrape(r.URL.String(), r.Method, r.Depth, r.Body, r.Ctx, *r.Headers, true)` in `Retry()`. It makes no sense to forbid URL revisit in `Retry()`, because `Retry()` is an explicit instruction to repeat a request.

The problem is that `Retry()` is used in `queue` incorrectly. I'd extend `Request` with a `Do()` or `Perform()` method which is similar to `Retry()` but respects `AllowURLRevisit` and I'd call this function from the `queue` instead of `Retry()`.
This way we can avoid API change. What do you think?

gsoec:
agree with you that `Retry()` should be an explicit instruction to repeat a request, but calling `Retry()` with a constant `checkRevisit = true` is always return immediately and doing nothing because URL is already visited. may be we could keep `r.collector.scrape(r.URL.String(), r.Method, r.Depth, r.Body, r.Ctx, *r.Headers, false)` inside `Retry` and change https://github.com/gocolly/colly/blob/d8c4e07fe2cc786ed5f42faeb350e9dac76edc7f/queue/queue.go#L150 to `r.collector.scrape(r.URL.String(), r.Method, r.Depth, r.Body, r.Ctx, *r.Headers, true)`,  
I think extending `Request` is nice approve to respect AllowURLRevisit.

asciimoo:
> but calling Retry() with a constant checkRevisit = true is always return immediately and doing nothing because URL is already visited

You are right.

> change to r.collector.scrape(r.URL.String(), r.Method, r.Depth, r.Body, r.Ctx, *r.Headers, true)

That's not possible, because `queue` is another package and `r.collector` is a private member in the `colly` package. That's why I suggested adding `Request.Do()` or `Request.Perform()` which would call `r.collector.scrape(r.URL.String(), r.Method, r.Depth, r.Body, r.Ctx, *r.Headers, !r.collector.AllowURLRevisit)`.

gsoec:
@asciimoo I have these changes

Issue #141, opened 2018-05-14
Suggestion: Ignore field tag for unmarshal func


ringxu666:
hello,
Is there a tag that spec the field should ignore when unmarshal from html just like json tag `json:"-"`

maybe like the code below?

```golang
func unmarshalAttr(s *goquery.Selection, attrV reflect.Value, attrT reflect.StructField) error {
	selector := attrT.Tag.Get("selector")
	if selector == "-" || selector == "" {
		return nil
	}
...
```

BTW, thanks for the great work . regards.

asciimoo:
Currently it is not supported. Could you open a PR with the mentioned additions?

ringxu666:
the embed struct define below in unmarshal_test should not be ignore. 
so only ignore when field tag specify as "-"
sorry for another pull request ^ ^
```golang
s := struct {
		String string   `selector:"li:first-child" attr:"class"`
		Items  []string `selector:"li"`
		Struct struct {
			String string `selector:"li:last-child"`
		}
}{}
```

PR #142, opened 2018-05-15
Add support to field unmarshal ignore specification

ringxu666:
#141


ringxu666:
only ignore when specification is "-".

PR #148, opened 2018-05-18
Convert http_backend to an interface

vosmith:
By converting the http_backend into a implementation of a common
interface, it allows flexibility of using other HTTP clients or
remote browsers to drive HTTP requests. This commit adds the HTTPDriver
interface which provides the same functionality as the http_backend and
updates the Collector to communicate the the http_backend through the
interface methods.

Also attempted to convert the AppEngine call to use the HTTPDriver interface

This is in support of issue #4

### DIFF
--- a/appengine_backend.go
+++ b/appengine_backend.go
@@ -103,14 +103,14 @@ func (h *appEngineBackend) Cache(request *http.Request, bodySize int, cacheDir s
 func (h *appEngineBackend) Do(request *http.Request, bodySize int) (*Response, error) {
 	r := h.GetMatchingRule(request.URL.Host)
 	if r != nil {
-		r.waitChan <- true
+		r.WaitChan <- true
 		defer func(r *LimitRule) {
 			randomDelay := time.Duration(0)
 			if r.RandomDelay != 0 {
 				randomDelay = time.Duration(rand.Intn(int(r.RandomDelay)))
 			}
 			time.Sleep(r.Delay + randomDelay)
-			<-r.waitChan
+			<-r.WaitChan
 		}(r)
 	}
 
--- a/appengine_backend.go
+++ b/appengine_backend.go
@@ -1,0 +1,207 @@
+// Copyright 2018 Adam Tauber
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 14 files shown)


appengine_backend.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+type appEngineBackend struct {
+	LimitRules []*LimitRule
+	Client     *http.Client
+	lock       *sync.RWMutex
+}
+
+func (h *appEngineBackend) Init(jar http.CookieJar) {
+	rand.Seed(time.Now().UnixNano())
```
manugarri: Is it necessary to set the seed here?

vosmith: :thinking:  probably not, it was a copy and paste from the current `http_backend`

manugarri: Good, I dunno if this has been fixed, but until a couple years ago [rand wasnt safe for concurrent use](https://blog.sgmansfield.com/2016/01/the-hidden-dangers-of-default-rand/) , not sure if it affects the seed

manugarri:
@asciimoo I assume that chromedp can trigger javascript (and thus is able to provide the raw html as you would see it in the browser). 

An easy way to check if this is correct (and make sure that chromedp is the right solution) is to try to scrape this [toy website](http://quotes.toscrape.com/). It requires javascript rendering in order to set up the dom the same way you see it in the browser.

asciimoo:
@manugarri I assume that too, but at some point we have to ask the response body from the browser. Then we have no option to interact with the browser.
What if I have to click a button to trigger an AJAX request (e.g. login) and I'm interested in the DOM after the AJAX request?

manugarri:
hmm good point, good point. @asciimoo , that would be equally important. Im a bit lost with chromedp to be honest, but it seems like colly would need to yield to the library in order to use chromedp api to interact with the page (because colly does not support page interaction)

vosmith:
Hey @asciimoo I've been playing around with chromdp and the HTTPDriver, and you're right, there would need to be some more modifications in order to have a meaningful relationship.

In order to get a copy of the HTML page, you can use `chromedp.OuterHTML("html", &somestring)`.

To answer the bigger question of interacting with the javascript, there would need to be a way were the `colly.Collector` can pass some information to the backend.  The cleanest solution I've been able to think of is through the `colly.Content`.  Right now the context is passed to `fetch` but not into `http_backend.Cache()`.  If we could do this then I could imagine passing simple actions to the chromedp, or any other driver.

asciimoo:
Another way to interract with the browser is extending the collector with an `OnBrowser` callback. Its' signature could be `func (c *Collector) OnBrowser(browserEvent string, func(*Browser))`, where `Browser` type would be the interface to the browsers, like `HTMLElement` in `OnHTML` callbacks.

vosmith:
@asciimoo I like the idea of having a Browser as a first class citizen.  I'm trying to think through what a Browser interface might look like.  I'll put some ideas up when I get some time to digest it.

vosmith:
I'm gonna have to move this to another PR.  I had rebase with master and now it all screwed up...

PR #152, opened 2018-05-25
i added random delay before request in queue

mazhigali:
and updated example queue

### DIFF
--- a/queue/queue.go
+++ b/queue/queue.go
@@ -28,7 +28,8 @@ type Storage interface {
 // Queue is a request queue which uses a Collector to consume
 // requests in multiple threads
 type Queue struct {
-	// Threads defines the number of consumer threads
+	// Threads defines the number of consumer threads.
+	//RandomDelay defines maximum time of delay before request
 	Threads           int
 	storage           Storage
 	activeThreadCount int32
@@ -56,7 +57,7 @@ type inMemoryQueueItem struct {
 
 // New creates a new queue with a Storage specified in argument
 // A standard InMemoryQueueStorage is used if Storage argument is nil.
-func New(threads int, randomDelay time.Duration, s Storage) (*Queue, error) {
+func New(threads int, s Storage) (*Queue, error) {
 	if s == nil {
 		s = &InMemoryQueueStorage{MaxSize: 100000}
 	}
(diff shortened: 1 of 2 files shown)


queue/queue.go
```diff
@@ -31,6 +33,7 @@ type Queue struct {
 	storage           Storage
 	activeThreadCount int32
 	threadChans       []chan bool
+	RandomDelay       time.Duration
```
asciimoo: Please add docstring to this new public member.

queue/queue.go
```diff
@@ -53,7 +56,7 @@ type inMemoryQueueItem struct {
 
 // New creates a new queue with a Storage specified in argument
 // A standard InMemoryQueueStorage is used if Storage argument is nil.
-func New(threads int, s Storage) (*Queue, error) {
+func New(threads int, randomDelay time.Duration, s Storage) (*Queue, error) {
```
asciimoo: `Queue.RandomDelay` is a public member of `Queue`, it isn't necessary to initialize `RandomDelay` here. Please remove the `randomDelay` argument to avoid API change.

queue/queue.go
```diff
@@ -147,6 +151,11 @@ func (q *Queue) Run(c *colly.Collector) error {
 					q.finish()
 					continue
 				}
+				randomDelay := time.Duration(0)
+				if q.RandomDelay != 0 {
+					randomDelay = time.Duration(rand.Intn(int(q.RandomDelay)))
```
asciimoo: You can call `sleep` in this line: `time.Sleep(time.Duration(rand.Intn(int(q.RandomDelay))))`

asciimoo:
I'm thinking on that this whole thing can be achieved using a single `OnRequest` callback and `sleep()`.
Has your solution any advantage compared to sleeping in an `OnRequest` callback?

mazhigali:
@asciimoo You are right that you can use OnRequest, but relying on the example http://go-colly.org/docs/examples/random_delay/ I thought that in the queue mechanism the c.Limit rules should work, but in practice it turned out not to be so. And it seemed logical to me to show that for the queuing mechanism, the query delay should be specified regardless of the settings of c.Limit

asciimoo:
`c.Limit` is a bit different, because it limits the access to specific resources (domains) defined by the user and not just waits after every request. `LimitRule` can be used with queues too, but if you want to sleep after every request, you have to set queue's `Parallelism` to `1`. I'd not pull this PR, because it does not add extra features to the framework that cannot be achieved without this PR.

Issue #155, opened 2018-05-29
Cannot auto guess encoding from html body on www.sdu.edu.cn


imfht:
[http://www.sdu.edu.cn](http://www.sdu.edu.cn) do not have encoding info in headers but has encoding info in html meta. colly cannot autodetect encoding type from response.
code snip
```golang
import (
	"github.com/gocolly/colly"
	"fmt"
)

func main() {
	c := colly.NewCollector(
		// Visit only domains: hackerspaces.org, wiki.hackerspaces.org
		//colly.AllowedDomains("hackerspaces.org", "wiki.hackerspaces.org"),
	)
	c.OnResponse(func(response *colly.Response) {
		fmt.Println(string(response.Body))
	})
	c.Visit("http://www.sdu.edu.cn")
}
```
get output
```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<meta http-equiv="Content-Type" content="text/html; charset=gb2312" />
<meta name="description" content="ɽ����ѧ,ɽ��,ɽ����ѧ�ٷ���վ,ɽ�����,SDU" />
<meta name="keywords" content="ɽ����ѧ,ɽ��,ɽ����ѧ�ٷ���վ,ɽ�����,SDU" />
<meta name="Copyright" content="Copyright (c) 2010 www.sdu.edu.cn All Rights Reserved." />
<title>��ӭ����ɽ����ѧ��ҳ</title>
<link href="../2010/images/style2.css" rel="stylesheet" type="text/css" />
<link href="../2010/images/ad2.css" rel="stylesheet" type="text/css" />
<script type="text/javascript" src="../2010/images/ad.js" ></script>
```
can colly auto get encoding from header or html body just like [scrapy](https://github.com/scrapy/w3lib/blob/ac1b7212b5f9badf2b35aaf6e094d839fb3ba858/w3lib/encoding.py#L48)?

asciimoo:
Try to use Colly's character detection heuristics: set `c.DetectCharset` to `true` before scraping. 

https://godoc.org/github.com/gocolly/colly#DetectCharset

PR #158, opened 2018-05-31
Get encoding from both body & header

imfht:
Hi:
    This PR is related to [this issue](https://github.com/gocolly/colly/issues/155). Auto get charset from body & header. Just like what scrapy and tika do.
    Why I need it? I'm doing common crawl use colly. Some website not return it's charset in header(such as http://www.sdu.edu.cn/). Colly could not auto guess it's encoding even when I add c.DetectCharset = true. In Another, the charset information has already in html meta. Ignore it and guess it by complex algorithm is not a good idea.
    I'm new to golang. hard for me to write a test example. you can close this pull request add write a more stable version ;-D


imfht:
ops, check failed.

PR #161, opened 2018-06-05
fix request produce after check empty stop program

mumugoah:
when produce request async, program maybe stop early


mumugoah:
wrong way to slove it

PR #171, opened 2018-06-21
Read gzipped response and request marshal include headers

### LINKED ISSUE #169: should request marshal include headers?
It will be difficult to deal with cookies and reference in headers
```golang
// Marshal serializes the Request
func (r *Request) Marshal() ([]byte, error) {
	ctx := make(map[string]interface{})
	if r.Ctx != nil {
		r.Ctx.ForEach(func(k string, v interface{}) interface{} {
			ctx[k] = v
			return nil
		})
	}
	var err error
	var body []byte
	if r.Body != nil {
		body, err = ioutil.ReadAll(r.Body)
		if err != nil {
			return nil, err
		}
	}
	return json.Marshal(&serializableRequest{
		URL:    r.URL.String(),
		Method: r.Method,
		Body:   body,
		ID:     r.ID,
		Ctx:    ctx,
	})
}
```

mumugoah:
1. add feather: read gzipped HTTP response
2. fix: request marshal and unMarshal include headers (fix #169 )

### DIFF
--- a/http_backend.go
+++ b/http_backend.go
@@ -28,6 +28,8 @@ import (
 	"sync"
 	"time"
 
+	"compress/gzip"
+
 	"github.com/gobwas/glob"
 )
 
@@ -184,6 +186,12 @@ func (h *httpBackend) Do(request *http.Request, bodySize int) (*Response, error)
 	var bodyReader io.Reader = res.Body
 	if bodySize > 0 {
 		bodyReader = io.LimitReader(bodyReader, int64(bodySize))
+	}
+	if res.Header.Get("Content-Encoding") == "gzip" {
+		bodyReader, err = gzip.NewReader(bodyReader)
+		if err != nil {
+			return nil, err
+		}
 	}
 	body, err := ioutil.ReadAll(bodyReader)
 	defer res.Body.Close()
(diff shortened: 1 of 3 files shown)


request.go
```diff
@@ -56,6 +56,7 @@ type serializableRequest struct {
 	Body   []byte
 	ID     uint32
 	Ctx    map[string]interface{}
+	Header http.Header
```
asciimoo: Please call it `Headers` to be compatible with the current naming convention.

http_backend.go
```diff
@@ -185,6 +187,12 @@ func (h *httpBackend) Do(request *http.Request, bodySize int) (*Response, error)
 	if bodySize > 0 {
 		bodyReader = io.LimitReader(bodyReader, int64(bodySize))
 	}
+	if res.Header.Get("Content-Encoding") == "gzip" {
```
asciimoo: As I know golang's http lib handles compression by default. If it successfully decompresses the content, then `res.Uncompressed` [1] is `true`, so we could check here if decompression happened to avoid double decompression:

```go
if !res.Uncompressed && res.Header.Get("Content-Encoding") == "gzip" {
```

[1] https://golang.org/pkg/net/http/#Response

mumugoah:
@asciimoo ok, I'll correct it

PR #182, opened 2018-07-04
Don't share wg field in Collector.Clone function

nange:


asciimoo:
What is the reason of this change? I think `wg` should be shared, because the new `collector` uses arrays of the original `collector` so both collectors have access to these arrays.

nange:
If we shared ```wg```, there's a problem.

For example:
```go
c1 := colly.NewCollector(colly.Async(true))
c2 := c1.Clone()
c3 := c1.Clone()

c1.OnHTML("selector1", func(e *colly.HTMLElement) {
	// do some thing, Visit link found on page
	c2.Visit("some url found in this page")
})

c2.OnHTML("selector2", func(e *colly.HTMLElement) {
	// do some thing, Visit link found on page
	c3.Visit("some url found in this page")
})

c3.OnHTML("selector3", func(e *colly.HTMLElement) {
	// some more logic to do...
        log.Println("some logic to do...")
})

for i := 0; i < 100; i++ {
        c1.Visit("http://xxx.xxx.com")
}

c1.Wait()     // if we shared wg field, when c1 completed, all of c2, c3 completed at the same time.
log.Println("c1 completed...")
c2.Wait()
log.Println("c2 completed...")
c3.Wait()
log.Println("c3 completed...")

```

If we shared wg field, when c1 completed, all of c2, c3 completed at the same time.  ```c2.Wait()``` and  ```c3.Wait()```become useless. That's not my expected.

asciimoo:
Ah, you're right, it's only necessary to share the lock. Thanks! =)

PR #188, opened 2018-07-10
Cache filter function

krzysztofantczak:
This PR should be rejected since it contains breaking changes in colly.CacheDir. However, i just wanted to demonstrate how it works at the moment.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -86,6 +86,9 @@ type Collector struct {
 	// CacheDir specifies a location where GET requests are cached as files.
 	// When it's not defined, caching is disabled.
 	CacheDir string
+	// CacheFilter specifies a function that filters response contents.
+	// Response should be cached only when function returns nil.
+	CacheFilter func(response *Response) error
 	// IgnoreRobotsTxt allows the Collector to ignore any restrictions set by
 	// the target host's robots.txt file.  See http://www.robotstxt.org/ for more
 	// information.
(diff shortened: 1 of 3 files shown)


asciimoo:
Useful functionality, thanks! Could you modify it to be an independent option from `CacheDir`? Then it wouldn't break the current API.

krzysztofantczak:
Something like colly.CacheDirWithFilter ?

krzysztofantczak:
Ok, i'm gonna need some help on dealing with my commits. My first commit on this PR kind of conflicts with latest one - i have tried squashing it, but i had to merge after that on my fork. How can i fix this in some nice way? I would go with fresh PR, but it seems like good learning moment ;-)

asciimoo:
Why do you want to use reflection for testing? I think, a simple test case would be enough:
 - Setup a custom `CacheFilter`
 - Create two requests where one is allowed by the `CacheFilter` and the other one isn't.
 - Check if the cache file created for the cached request.

Another thought:
It would be good to support every type of requests, not only GET. We can achieve it if we define a default `CacheFilter` function for a collector instead of doing additional filtering in `httpBackend.Cache`.

vosmith:
@krzysztofantczak are you able to continue this PR?  Looks like good progress was made :+1:

Splik:
@krzysztofantczak, this feature is really appreciable one, so +1 to @vosmith.

asciimoo:
Closing due to inactivity.

PR #12, opened 2017-10-06
Add a dependency 

NafisFaysal:
close issue #11

PR #198, opened 2018-07-29
Include missing package in vendor

nafisfaysal:
Here the story: #12


asciimoo:
@nafisfaysal how does this fix the broken dependency issue mentioned in #12?

nafisfaysal:
@asciimoo If I got you right, I already mention there why I like to keep vendor folder in the library. FYI I put here my previous comment and `golang-nuts` thread about [Vendoring](https://groups.google.com/forum/#!topic/golang-nuts/yASFt3PY_yg) 


> This PR has been almost nine months. Recently I experimented more Golang and Golang stuff. I think Vendoring is one of the rights approached to keep the library more securing from failing. Like colly use some third-party libraries if those libraries Author deletes the repository colly will be failed to work properly Right. Here of some awesome work https://github.com/docker/cli https://github.com/dgraph-io/dgraph they push the vendor folder and keep it and if anytime needs to update it they will.

> So, my point it I think @asciimoo can keep the vendor folder for the colly and @zet4 thanks for the comment to keep me busy to experiment more about the dependency stuff.

> Note: It's my personal opinion. As always, feedback (be it constructive, or destructive) are welcome.

Anyway, keep the vendor folder on the project repo depends on the project Author. As always I welcome and respect the community resolution to make a better ecosystem for the Go community.

asciimoo:
@nafisfaysal I like the idea of vendoring, but there was a problem with the previous vendoring solution. Colly exposes the `goquery` go package, and if we vendor the packages, then users cannot use their `goquery` package, because colly imports a "different" vendored one and this leads to a compile error, as we can see on the image from the previous thread:

That's why I reverted the previous vendoring attempt. Do you know how can we avoid this situation with any kind of vendoring?

PR #203, opened 2018-08-07
update `Collector.Appengine` to take a `context.Context` rather than an `*http.Request`

peterellisjones:
The convention for Google App Engine applications is to pass around a `context.Context` rather than an `*http.Request`. Typically you pass context.Context around your application and use it to call various GAE services. This update means you can call Colly deep in your code somewhere where you only have `context.Context` available as you'd typically only have `*http.Request` available near the ingress point.

I have confirmed these changes work in a fork I am using on a live GAE app.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -17,6 +17,7 @@ package colly
 
 import (
 	"bytes"
+	"context"
 	"crypto/rand"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


asciimoo:
Nice! This PR changes the API of Colly so if we want to follow semantic versioning, this change would increase Colly version to `2.0.0`. I'm thinking on bumping to `1.1.0` after the merge, because this isn't a huge change and doesn't affect the core functionality of Colly. What do you think?

peterellisjones:
Hi @asciimoo — thanks for the taking a look at this.

> I'm thinking on bumping to 1.1.0 after the merge, because this isn't a huge change and doesn't affect the core functionality of Colly. What do you think?

That sounds reasonable to me.

asciimoo:
Great, thanks for the contribution!

Issue #110, opened 2018-03-07
Get proxy URL for request


matthew-noken:
I'd like to know which proxy URL was used with which request so that on errors I can automatically remove proxies that return repeated errors or timeouts. Ideally it would be a property, like Request.ProxyURL. It doesn't look like there's any way to do that currently, though.

I'm happy to contribute a PR, but I'm not seeing an obvious way to get at that except by keeping a map[string]string ledger which is less than ideal. Thoughts?

asciimoo:
@matthew-noken yes, this is definitely a valid use-case and currently it isn't possible to retrieve the proxy URL in callbacks.
I'd extend the `Request` struct with a `Proxy() string` member function, because `Request` is accessible from all the callbacks. What do you think?

matthew-noken:
I think that's good, but I'm trying to figure out how to get at the proxy URL for a particular request.

The call flow looks like this:

1. User passes a ProxyFunc (such as the return value of proxy.RoundRobinProxySwitcher()) to colly.Collector.SetProxyFunc()
2. colly.Collector.SetProxyFunc() assigns ProxyFunc to http.Transport.Proxy
3. http.Transport.RoundTrip() calls http.Transport.connectMethodForRequest()
4. http.Transport.connectMethodForRequest() calls http.Transport.Proxy() (our ProxyFunc).

The return value of connectMethodForRequest(), though—the proxy URL itself—is ephemeral. It lives and dies in http.Transport.RoundTrip().

Ultimately the proxy URL used needs to be assigned to a Request somewhere, so the question is where? As I see it, because the request is passed to the ProxyFunc, the best option is to embed http.Request in colly.Request and pass it to http.Client directly. Add a ProxyURL property. Modify proxy.roundRobinSwitcher.GetProxy to set colly.Request.ProxyURL (and change the name to something like SelectProxyURL, since it has side effects). This can retain public interface backwards compatibility.

I can create a PR if this sounds good.

asciimoo:
@matthew-noken you are right, this solution seems valid to me. I'd appreciate a PR. Thanks!

matthew-noken:
(I've been super busy, but I do still plan to do this in the next 2-3 weeks.)

webzak:
It is possible to set additional request header in ProxyFunc like:
```
func ProxyFunc(r *http.Request) (*url.URL, error) {
        proxy := "socks5://127.0.0.1:1777"
	r.Header["proxy"] = []string{proxy}
[...]

PR #222, opened 2018-09-17
Proxy URL for request

pyjac:
- Add ProxyURL to Request
- Add ProxyURL via context
- Update example

Addresses #110

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -158,6 +158,13 @@ type cookieJarSerializer struct {
 
 var collectorCounter uint32
 
+// The key type is unexported to prevent collisions with context keys defined in
+// other packages.
+type key int
+
+// ProxyURLKey is the context key for the request proxy address.
+const ProxyURLKey key = 0
+
 var (
 	// ErrForbiddenDomain is the error thrown if visiting
 	// a domain which is not allowed in AllowedDomains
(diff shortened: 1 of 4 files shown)


proxy/proxy.go
```diff
@@ -27,9 +28,11 @@ type roundRobinSwitcher struct {
 	index     uint32
 }
 
-func (r *roundRobinSwitcher) GetProxy(_ *http.Request) (*url.URL, error) {
+func (r *roundRobinSwitcher) GetProxy(pr *http.Request) (*url.URL, error) {
 	u := r.proxyURLs[r.index%uint32(len(r.proxyURLs))]
 	atomic.AddUint32(&r.index, 1)
+	ctx := context.WithValue(pr.Context(), colly.ProxyURLKey, u.String())
+	*pr = *pr.WithContext(ctx)
```
vosmith: Because `WithContext()` already makes a copy of `pr`, I think its better to do this without any dereferencing.

colly.go
```diff
@@ -158,6 +158,13 @@ type cookieJarSerializer struct {
 
 var collectorCounter uint32
 
+// The key type is unexported to prevent collisions with context keys defined in
+// other packages.
+type key int
+
+// ProxyURLKey is the context key for the request proxy address.
+const ProxyURLKey key = 0
```
vosmith: I think it would be better to stick with the `iota` here instead of creating a new type.

proxy/proxy.go
pyjac: Without doing this, I can't think of another way to update the `Context` of `pr`. Just like you said, `WithContext` only makes a copy and does not update the `Context` of `pr`.

mazhigali:
I think you should close the issue, because it is done in https://github.com/gocolly/colly/blob/master/request.go#L52

Issue #223, opened 2018-09-19
Handler on html not working if response header not contain Content-Type (response is html)


wforfun:
Sorry for any mistake, English is not my native language. 
As a title, some website not set Content-Type on response header so  function handleOnHTML return nil when validate response type is html.
```
func (c *Collector) handleOnHTML(resp *Response) error {
	if len(c.htmlCallbacks) == 0 || !strings.Contains(strings.ToLower(resp.Headers.Get("Content-Type")), "html") {
		return nil
	}
        ....
}
```
I try to use body response to validate it is html,  looks like it works 

```
func (c *Collector) handleOnHTML(resp *Response) error {
	if len(c.htmlCallbacks) == 0 || !isResposeHTML(resp) { {
		return nil
	}
        ....
}

func isResposeHTML(resp *Response) bool {
	fmt.Println(strings.ToLower(string(resp.Body)[0:15]))
	if strings.ToLower(string(resp.Body)[0:15]) == "<!doctype html" {
		return true
	}
	return false
}
```

PR #224, opened 2018-09-19
fix issues #223 

pinkykitten:
https://github.com/gocolly/colly/issues/223
Use body response to validate it is html instead of use response header Content-type

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1281,7 +1281,7 @@ func isMatchingFilter(fs []*regexp.Regexp, d []byte) bool {
 }
 
 func isResposeHTML(resp *Response) bool {
-	if strings.ToLower(string(resp.Body)[0:15]) == "<!doctype html>" {
+	if strings.ToLower(string(resp.Body)[0:14]) == "<!doctype html" {
 		return true
 	}
 	return false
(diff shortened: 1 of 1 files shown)


asciimoo:
Detecting the content type this way leads to errors too if the content doesn't start with '<!doctype html>`.
E.g. https://www.imdb.com has an extra whitespace before the docstring declaration.

The proper solution would be a validator public member function which could handle non-standard edge cases too. Something like `Collector.IsHTML func(Response) bool`. What do you think?

wforfun:
Collector. IsHTML func (Response) bool``` 
Agree, can you give me some suggestion? 
Can I getting first 200 character of the body, then confirm it contain `doctype html`

vosmith:
I think checking the `doctype` tag is the right approach, but perhaps the body should be passed through `strings.TrimSpace()` first.  I think any characters other than whitespace would break html parsing any ways.

Or perhaps use `regex` but I'm not sure which would be more efficient.

asciimoo:
These are just heuristics, so I'd like to be able to change it, that's why I suggested a public attribute for this function. I can have a textfile which starts with `<!doctype`, this case it would be good to overwrite the default heuristics function (which checks the content-type and/or the first n bytes of the response).

asciimoo:
Closing due to inactivity.

PR #225, opened 2018-09-20
:sparkles: Add get raw HTML element

faruqisan:
Changes : 

- Add HTML field in `HTMLElement` struct that contain raw html element given

Notes : 

- No error test coverage on 
```
html, err := goquery.NewDocumentFromNode(n).Html()
	if err != nil {
		log.Println("err on get html from node, ", err)
	}
`

### DIFF
--- a/htmlelement.go
+++ b/htmlelement.go
@@ -15,6 +15,7 @@
 package colly
 
 import (
+	"log"
 	"strings"
 
 	"github.com/PuerkitoBio/goquery"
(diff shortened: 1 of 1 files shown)


asciimoo:
Thanks for your contribution.
You can use `HTMLElement.DOM.Html()` to get the selection's raw HTML content. It has error handling and it is more efficient (it doesn't allocate memory for the HTML content in every element). As I see this is a redundant functionality so closing.

faruqisan:
I'm sorry that I didn't know that i can use `HTMLElement.DOM.Html()` before. that's why i try to create this PR. any way thanks for your response @asciimoo :+1:

PR #230, opened 2018-10-01
Fixes issue #229

### LINKED ISSUE #229: Referer extension : wrong header
In the Referer extension, the HTTP Header should be "Refe**r**er" instead of "Refe**rr**er".

(btw : https://en.wikipedia.org/wiki/HTTP_referer#Etymology)

bmancone:


asciimoo:
Not exactly sure why, but travis started to fail: https://travis-ci.org/gocolly/colly/jobs/435581027

bmancone:
@asciimoo it looks like the master branch also fails, right ?

PR #246, opened 2018-10-27
Increment i before return True

nigelbabu:
The counter does not get incremented at all at the moment


nigelbabu:
There doesn't seem to be any tests for `htmlelement.go`, so I haven't added one. If I have somehow missed them, please let me know.

PR #245, opened 2018-10-27
Added Go module support

Zaba505:

Issue #240, opened 2018-10-24
context.Context Support


Zaba505:
# Question
I was wondering if it'd be possible to add support for wrapping the On* callbacks with a context or possibly adding a context to the callback function signature. This can most likely be achieved with little to no overhead by adopting the pattern of having two calls one with context and one without. e.g. `OnHtml()` and `OnHTMLWithContext()`.

# Reason
The reason for implementing this is for concurrent models which seperate the data processing from the data ingestion. A simple example of this would be using colly to scrape Facebook posts and then send them on a channel to some goroutine that's maybe doing some real-time analytics on them. Implementing this model currently with colly requires you to know if the channel is closed or not before sending. There is also no way to stop colly's execution either so it will keep scraping even though the goroutine may have already stopped processing due to some error.

# Possible API
The key to this is having an explicit API to facilitate the execution model with a context. After putting some thought into this (as I was writing this), I believe valid implementations for this are either:
1) A method, `VisitWithContext()`, could be added
or
2) A `context.Context` field will have to added to the `colly.Collector` type and can be set with an option function

Both of these solutions can be implemented together; however, I think it's best to keep the API as simple as possible and as explicit as possible. In my opinion, option 1 seems to be the better route since it's extremely explicit, whereas implementing option 2 would require option 1 as well, in order to achieve the same expressiveness in the API. I also think that to avoid having to create more `colly.Collector` fields for callbacks with contexts, it may be better to just completely rewrite the callback signature to include a `context.Context` by default. This would then follow in the footsteps of other APIs like gRPC, which includes a context in every RPC signature.

vosmith:
Hi @Zaba505 are you looking to use context.Context in order to Cancel() them in the case that a channel closes? 

If you are just holding key-value data, you can capture the request with an onRequest callback, add your information to that request and then in your onHTML you can access the HTMLElement.Request.Ctx there.

Am I on the right track here?

Zaba505:
[...]

PR #255, opened 2018-11-15
Added context.Context Support

Zaba505:
This extends PR #245 by adding context.Context support, as discussed in Issue #240. This contribution comes after discussion and example implementations from both myself and @vintik100.


fooofei:
I support a less example code to use `context.Context` with colly,

which not change anything in colly.

```golang
// contextTransport wrapper a context.Context for cancel requests
type contextTransport struct {
	ctx   context.Context
	trans *http.Transport
}

func (t *contextTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	req = req.WithContext(t.ctx)
	return t.trans.RoundTrip(req)
}

func collectorWithContext(c *colly.Collector, ctx context.Context) {
	// We can stop all requests at `OnRequest` callback 
    // before send request to HTTP client.
    c.OnRequest(func(req *colly.Request) {
		select {
		case <-ctx.Done():
			req.Abort()
		default:
		}
	})

    // Use custome Transport to cancel all pending requests at HTTP client,
    // which not have chance to stop at OnRequest callback.
    trans := &contextTransport{
		ctx:   ctx,
		trans: &http.Transport{},
	}
	c.WithTransport(trans)
}
```

This will work when cancel the `context.Context`,

`ctx.Done()` will return, and all pending request will immediately be canceled.

WGH-:
Hey! Any reason why this became stuck?

Is it because this PR changes lots of function signatures to have context argument, which is a pretty radical change?

If so, what about making it much less drastic by only adding `Collector.WithContext` method, which will be used for the `http.Request`? This way API compatibility would be kept, and it still would be possible to cancel HTTP requests in a clean way.

And since the common way to use colly is to have `OnHTML` etc. handlers be closures, you can simply use your own context from the closure. E.g.

```go
func crawl(ctx context.Context, url string) {
	c := colly.NewCollector(colly.WithContext(ctx))
	c.OnResponse(func(res *colly.Response) {
		storeIntoSomeDatabase(ctx, res)
	})
	// skip
}

asciimoo:
> If so, what about making it much less drastic by only adding Collector.WithContext

I like this approach a lot! @WGH- could you work on this?

WGH-:
@asciimoo yes

FWIW, I said it in other issue, but the solution outlined by @fooofei is not perfect since it loses the original request's context which might already have deadline set. See `net/http/client.go`, where `setRequestCancel(req, rt, deadline)` is called just before `RoundTrip`.

asciimoo:
> FWIW, I said it in other issue, but the solution outlined by @fooofei is not perfect since it loses the original request's context which might already have deadline set

I think it isn't a deal breaker. We can document this behavior or perhaps we can throw a warning if the user sets a custom timeout and also uses custom ctx. What do you think?

WGH-:
> > FWIW, I said it in other issue, but the solution outlined by @fooofei is not perfect since it loses the original request's context which might already have deadline set
> 
> I think it isn't a deal breaker. We can document this behavior or perhaps we can throw a warning if the user sets a custom timeout and also uses custom ctx. What do you think?

What I said earlier applies only to the hack with setting context inside `RoundTrip`. Setting context before passing the `http.Request` to `Client.Do` does not have this problem.

StantonMatt:
I think this one is obsolete now.

Current `master` already has the smaller approach discussed above, under the name `StdlibContext(ctx)`: it stores the stdlib context on the collector and `scrape` applies it with `req.WithContext(...)` before the request goes through `Client.Do`. The committed `TestCollectorContext` passes on current `master`, and a local canceled-context probe also returned `context.Canceled` without reaching the test server.

The submitted branch still passes its old `go test ./... -count=1`, but it is no longer close to mergeable: `git merge-tree origin/master HEAD` reports conflicts across `colly.go`, `context.go`, `http_backend.go`, `request.go`, several examples, and `go.mod`/`go.sum`. It also still changes the public callback signatures (`OnRequest`, `OnResponse`, `OnHTML`, `OnXML`, `OnError`, `OnScraped`), so current callback registrations no longer compile.

Given that current `master` has the API-compatible context support, I don't think this branch is worth rebasing.

PR #277, opened 2019-01-26
Add SetClient to set custom http client

geertweening:
So we can use http client and cookie jar across collectors and outside collector

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -837,6 +837,11 @@ func (c *Collector) OnScraped(f ScrapedCallback) {
 	c.lock.Unlock()
 }
 
+// SetClient will override the previously set http.Client
+func (c *Collector) SetClient(client *http.Client) {
+	c.backend.Client = client
+}
+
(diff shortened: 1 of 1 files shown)


estroz:
@asciimoo any chance of this getting merged soon?

asciimoo:
I'm not sure if this is the right approach. If the goal is to get the client and the cookie jar, then a `GetClient` and `GetCookieJar` getter is enough. What is the benefit of allowing to set a custom http client?

estroz:
@asciimoo custom testing of a scraper built with colly is an example:

```go
func TestSomeFunction(t *testing.T) {
	s := httptest.NewTLSServer(newTestHandler())
    defer s.Close()
    collector := colly.NewCollector()
    collector.SetClient(s.Client())
	...
}
```

Or if you have other custom client settings.

estroz:
@asciimoo what do you think about my comment above? Apologies for bugging you about this, I really like the project and appreciate your contributions to OSS!

PR #296, opened 2019-02-26
Fix runtime error when proxy is empty

JohnZhao1208:
```golang
rp, err := proxy.RoundRobinProxySwitcher()
if err != nil {
	log.Fatal(err)
}
```

Code like this will cause a runtime error ,
try to fix this.


vosmith:
Hey @JohnZhao1208 , Thanks for this PR!  Looks like a good fix for this situation.  Could you please add a test case to cover your new code?

Issue #317, opened 2019-03-27
OnResponse is not called when response has empty body


icamys:
Let's take a domain - `webmaster.immobilien-expo.de`

curl output:

```
curl -v 'webmaster.immobilien-expo.de'

*   Trying 212.114.63.124...
* TCP_NODELAY set
* Connected to webmaster.immobilien-expo.de (212.114.63.124) port 80 (#0)
> GET / HTTP/1.1
> Host: webmaster.immobilien-expo.de
> User-Agent: curl/7.63.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: nginx/1.14.1
< Date: Wed, 27 Mar 2019 11:53:35 GMT
< Content-Type: text/html
< Content-Length: 0
< Connection: keep-alive
< X-Powered-By: PHP/5.4.45
< 
* Connection #0 to host webmaster.immobilien-expo.de left intact
```

When we take this url and try to parse it with colly, the only called handler is OnRequest. Neither OnResponse nor OnError are called. Unfortunately there is nothing like OnFinished handler mentioned in #305.

icamys:
Looks like the reason is that in [response.go:85](https://github.com/gocolly/colly/blob/9192be95497e67fcc0f412347fe836c9e505d4ea/response.go#L85) the variable r.Body has zero length, function encodeBytes returns EOF error which is returned to fetch() and as a result, handleOnResponse is not called.

asciimoo:
Thanks for the fix (#318).

PR #318, opened 2019-03-27
Do not fix charset for empty body (fix for #317)

icamys:


icamys:
@asciimoo Review it please

PR #358, opened 2019-07-29
Feature/deadline

derricw:
Fixed a potentially indefinite hang when collecting data from a website that requests an upgrade to websocket.

Accomplished by adding a deadline to the response body read.


derricw:
Didn't mean to open this upstream yet. I meant to do it on my fork for a while to test.

PR #363, opened 2019-08-12
Add Go module support

twpayne:
This is a fresh generation of `go.mod` and `go.sum`, like #245, but without the unrelated code changes.


sky126:
module github.com/gocolly/colly/v2

PR #408, opened 2019-12-01
Cleanup invalid xml characters before unmarshalling

jredl-va:
@asciimoo for your consideration here is another PR related to cleanup of xml sitemaps. When unmarshalling xml in go I've run into several sites with bad unicode characters:

```
2019/11/30 18:29:56 Error from https://somesite.com/somethin.xml with status code 200: XML syntax error on line 1: illegal character code U+0003
```

This will cleanup the file before passing onto the xml parser.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -37,6 +37,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unicode"
 
 	"github.com/PuerkitoBio/goquery"
 	"github.com/antchfx/htmlquery"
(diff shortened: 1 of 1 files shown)


StantonMatt (commented):
I checked this one against current `origin/master`. The reported failure still reproduces: a local regression with `<page><title>before\x03after</title></page>` returns `XML syntax error on line 1: illegal character code U+0003` on master.

The submitted branch fixes that specific case, but it also strips XML-valid whitespace because it uses `unicode.IsPrint`. A second local probe with `<title>a\nb\tc\rd</title>` passes on master, but on this branch the callback sees `abcd` instead of preserving the newline/tab/CR-normalized text.

A narrower current-master scratch patch that only drops XML-invalid runes (`\t`, `\n`, `\r`, `0x20..0xD7FF`, `0xE000..0xFFFD`, `0x10000..0x10FFFF` kept) passed both probes, the existing XML tests, `go test ./... -count=1`, and `git diff --check origin/master` locally. So I think the bug is real, but this should be rebased with an XML-character-validity filter rather than a printability filter.

PR #419, opened 2019-12-16
add Statistics function to colly.Collector

makelove:
Statistics function is very useful

we can upload Statistics data to database, then monitor it

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -752,6 +752,18 @@ func (c *Collector) String() string {
 	)
 }
 
+//return Statistics in map
+func (c *Collector) Statistics() map[string]int {
+	return map[string]int{
+		"RequestCount": int(c.requestCount),
+		"ResponseCount": int(c.responseCount),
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Thanks!

asciimoo:
Please fix gofmt/golint errors.

asciimoo (changes requested):
Please fix code style errors.

niedbalski:
It would be very nice if this can be merged @asciimoo and added into a prometheus exporter :-)

StantonMatt (commented):
I checked the current head `fe0f649` against `origin/master`.

These passed locally:

```text
go test ./... -count=1
```

The branch still needs the style cleanup that was requested earlier, though. `gofmt -l colly.go` still reports:

```text
colly.go
```

The exported method comment is also currently `//return Statistics in map`; after formatting, I would make that a normal exported-symbol comment such as `// Statistics returns ...` so the public API doc reads cleanly.

PR #420, opened 2019-12-18
Update func isDomainAllowed

jiguangsdf:
like scrapy. if c.Visit("http://www.xxxx.com.cn/"), colly.AllowedDomains("xxxx.com.cn") should be work for main domain. so if we request 'www.xxxx.com.cn' and AllowedDomains are 'xxxx.com.cn'. This means that we have access to all subdomains under the top-level domain name

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -694,6 +694,11 @@ func (c *Collector) isDomainAllowed(domain string) bool {
 		return true
 	}
 	for _, d2 := range c.AllowedDomains {
+		if strings.Contains(domain, d2) {
+			return true
+		}
+	}
+	for _, d2 := range c.AllowedDomains {
 		if d2 == domain {
 			return true
 		}


colly.go
```diff
@@ -693,6 +693,11 @@ func (c *Collector) isDomainAllowed(domain string) bool {
 	if c.AllowedDomains == nil || len(c.AllowedDomains) == 0 {
 		return true
 	}
+	for _, d2 := range c.AllowedDomains {
+		if strings.Contains(domain, d2) {
```
belintani: I recommend to use "HasSuffix" method instead of "Contains". Does that make sense?

maaronking:
This is a great change, I was just looking for a good solution for this issue. 👍

colly.go
jiguangin: yeah. "HasSuffix" is well

asciimoo: HasSuffix sounds better for me, could you change it? Also, please update the docstring with the new behavior.

edoardottt:
Any update about this? ...

WGH-:
I dunno, this is a pretty breaking change. Depending on your use case, you might want this behaviour, or not.

In any case, one can do arbitrary origin check of the URL just before calling `Visit`, having it built in into `Collector` is not strictly necessary.

StantonMatt (changes requested):
I checked the current PR head `a8221a2` against current `master`.

The submitted branch still passes the existing suite:

```text
go test . -run TestCollectorVisitWithAllowedDomains -count=1 -v
go test ./... -count=1
```

The main blocker is that the new `strings.Contains(domain, allowedDomain)` check makes the allowed-domain boundary too broad. With `AllowedDomains("example.com")`, this branch also allows unrelated hosts that merely contain that string, such as `badexample.com` and `example.com.evil`.

I verified that with a local-only guard test:

```text
AllowedDomains(example.com) allowed unrelated host "badexample.com"
```

The same guard passes on current `master`, which still requires an exact allowed-domain match. So this would expand the crawl scope beyond the configured domain list.

Two current-state notes: `git merge-tree origin/master a8221a2` reports a conflict in `colly.go`, and `gofmt -l colly.go` reports `colly.go` on the PR branch.

PR #424, opened 2019-12-25
Add OnResponseHeaders callback

### LINKED ISSUE #228: OnResponseHeader
I ran into a problem where I am downloading a whole bunch of PDF files. It would be nice have a callback where I can parse the response headers before proceeding with the rest of the body and drop the request based on mime type.

I realize I could probably use the HEAD method as a work around, but some servers / web frameworks aren't always up to spec in supporting it properly.

So I think an on response header hook would be a great enhancement. I'm sure there would be more than a few use-cases outside my own for it.

WGH-:
The callback is called just before reading the response body, and can be used to abort the request.

Closes #228.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -109,25 +109,29 @@ type Collector struct {
 	CheckHead bool
 	// TraceHTTP enables capturing and reporting request performance for crawler tuning.
 	// When set to true, the Response.Trace will be filled in with an HTTPTrace object.
-	TraceHTTP         bool
-	store             storage.Storage
-	debugger          debug.Debugger
-	robotsMap         map[string]*robotstxt.RobotsData
-	htmlCallbacks     []*htmlCallbackContainer
-	xmlCallbacks      []*xmlCallbackContainer
-	requestCallbacks  []RequestCallback
-	responseCallbacks []ResponseCallback
-	errorCallbacks    []ErrorCallback
-	scrapedCallbacks  []ScrapedCallback
-	requestCount      uint32
-	responseCount     uint32
-	backend           *httpBackend
-	wg                *sync.WaitGroup
-	lock              *sync.RWMutex
+	TraceHTTP                bool
+	store                    storage.Storage
+	debugger                 debug.Debugger
+	robotsMap                map[string]*robotstxt.RobotsData
+	htmlCallbacks            []*htmlCallbackContainer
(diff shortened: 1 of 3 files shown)


LeMoussel:
Very interesting 💯

LeMoussel:
No commit in V 2.0. Damage.

colly.go
```diff
@@ -768,6 +779,19 @@ func (c *Collector) OnRequest(f RequestCallback) {
 	c.lock.Unlock()
 }
 
+// OnResponseHeaders registers a function. Function will be executed on every response
+// when headers and status are already received, but body is not yet read.
+//
+// Like in OnRequest, you can call Request.Abort to abort the transfer. This might be
+// useful if, for example, you're following all hyperlinks, but want to avoid
+// downloading files. Be aware that using this will prevent HTTP/1.1 connection reuse,
```
asciimoo: Does it prevent connection reuse only in case of abort?

WGH-: Yes.

WGH-: Though I didn't test HTTP/2.0, _I think it should_ handle it gracefully.

asciimoo: Great, thanks, could you update the PR to the latest master? And it's ready to merge

WGH-: I checked HTTP/2 closedly, just in case. We indeed don't lose the connection reuse in HTTP/2 case. When `Response.Body` is closed, a single stream is reset, and connection remains alive: https://github.com/golang/net/blob/6afb5195e5aab057fda82e27171243402346b0ad/http2/transport.go#L2077-L2104. I verified it in Wireshark as well.

WGH-: > Great, thanks, could you update the PR to the latest master? And it's ready to merge

Yes, in a couple of days.

asciimoo: Awesome, thanks!

WGH-:
I rebased the PR on current master, and added a couple of clarifying comments.

asciimoo:
Great, thanks, travis is still failing, could you fix it?

WGH-:
I don't think so: so failure doesn't seem to be caused by my changes.

asciimoo:
You are right, sorry, a job restart solved the problem. Thanks again for your contribution.

MeNsaaH:
@asciimoo, when will the new version with this change be released?

PR #425, opened 2019-12-25
Move isDomainAllowed and robots.txt checks inside requestCheck

WGH-:
Previously, these checks were left outside the `requestCheck` function, and run after it.

Since the `requestCheck` has a side effect of marking URL as visited, URLs filtered out by DisallowedDomains blacklist or robots.txt would return `ErrForbiddenDomain`/`ErrRobotsTxtBlocked` on first visit attempt, but `ErrAlreadyVisited` on subsequent revisit attempts, which is inconsistent with behaviour of other checks.


LeMoussel:
It would be interesting that these functions are accessible (`IsDomainAllowed()` & `CheckRobots()`)

WGH-:
@LeMoussel sorry, I dont follow

PR #426, opened 2019-12-25
Httptrace

tsvtitan:
Sometimes there is a need to to track every request DNS, TLSHandshake and Connect duration based on https://golang.org/pkg/net/http/httptrace/.


tsvtitan:
/v2 removed cause based on go.mod it should be in /v2 subfolder in repo

PR #434, opened 2020-01-16
Add OnJSON callback function

490694561:
When I was crawling the Github API, I found that I needed to repackage the parsing of dynamic json for OnResponse, so I wanted to submit an OnJSON function

Using Example :

```golang
c.OnJSON(func(jsonT []map[string]interface{}) {
	for _, row := range jsonT {
		log.Println(row["login"])
	}
})
```

> I am new to Golang, if the code is not good, please forgive me, tell me, I will modify it

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -198,7 +198,8 @@ var (
 	// ErrEmptyProxyURL is the error type for empty Proxy URL list
 	ErrEmptyProxyURL = errors.New("Proxy URL list is empty")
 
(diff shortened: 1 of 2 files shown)


490694561:
Somebody else?

asciimoo:
What are the benefits of this callback?

490694561:
I just think that it will be more convenient for me to crawl the API. If nothing else, I will close the problem myself.

490694561:
Maybe next time I will learn to issue first, maybe not next time

PR #436, opened 2020-01-17
Store responses/requests dumps

CorentinB:
Hi!

I am working on a WARC file writing collector for Colly, to write proper WARC files I need the most authentic requests and responses content to be saved, the best I've found was to use [DumpResponse](https://golang.org/pkg/net/http/httputil/#DumpResponse) and [DumpRequestOut](https://golang.org/pkg/net/http/httputil/#DumpRequestOut) to dump the content of the original http.Response and http.Request, Colly doesn't expose these 2 structures via colly.Response and colly.Request, so I made a patch that expose the dumps of the original `net/http` structures.

Thanks!

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -28,7 +28,6 @@ import (
 	"log"
 	"net/http"
 	"net/http/cookiejar"
-	"net/http/httputil"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -110,10 +109,7 @@ type Collector struct {
 	CheckHead bool
 	// TraceHTTP enables capturing and reporting request performance for crawler tuning.
 	// When set to true, the Response.Trace will be filled in with an HTTPTrace object.
-	TraceHTTP bool
-	// Dump enables dumping of requests and responses,
-	// dumps are stored in Request.Dump and Response.Dump.
-	Dump                     bool
+	TraceHTTP                bool
 	store                    storage.Storage
 	debugger                 debug.Debugger
 	robotsMap                map[string]*robotstxt.RobotsData
@@ -252,9 +248,6 @@ var envMap = map[string]func(*Collector, string){
 	"TRACE_HTTP": func(c *Collector, val string) {
 		c.TraceHTTP = isYesString(val)
 	},
-	"DUMP": func(c *Collector, val string) {
-		c.Dump = isYesString(val)
-	},
 	"USER_AGENT": func(c *Collector, val string) {
 		c.UserAgent = val
 	},
@@ -362,13 +355,6 @@ func TraceHTTP() CollectorOption {
 	}
 }
 
-// Dump instructs the Collector to generate dumps of requests and responses
-func Dump() CollectorOption {
-	return func(c *Collector) {
-		c.Dump = true
-	}
-}
-
 // ID sets the unique identifier of the Collector.
 func ID(id uint32) CollectorOption {
 	return func(c *Collector) {
@@ -417,7 +403,6 @@ func (c *Collector) Init() {
 	c.IgnoreRobotsTxt = true
 	c.ID = atomic.AddUint32(&collectorCounter, 1)
 	c.TraceHTTP = false
-	c.Dump = false
 }
 
 // Appengine will replace the Collector's backend http.Client
@@ -617,7 +602,6 @@ func (c *Collector) fetch(u, method string, depth int, requestData io.Reader, ct
 	if ctx == nil {
 		ctx = NewContext()
 	}
-
 	request := &Request{
 		URL:       req.URL,
 		Headers:   &req.Header,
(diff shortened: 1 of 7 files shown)


asciimoo (changes requested):
This is a good feature but the implementation requires a lot of memory to store the duplicate of the full response. Could you change the code to generate this dump by an individual function of `Request`/`Response` objects that we can call in case we need it. It would allow us to reduce memory usage if we don't use these dumps.

Also, please add tests.

CorentinB:
@asciimoo thanks for your review! Are you talking about something like that (let's take the Response dump as the example):

```
// Dump return the dump of the Response
func (r *Response) Dump() (dump []byte, err error) {
	dump, err = httputil.DumpRequestOut(HTTP_RESPONSE_GOES_HERE, true)
	if err != nil {
		return dump, err
	}

	return dump, nil
}
```

If yes, then we would need to store a copy of the original `net/http` response struct to do this kind of things, because I need to access it. HTTP_RESPONSE_GOES_HERE could be something r.OriginalResponse if we do that.

dvanduzer:
It would be great to find a way to reuse the existing memory allocations for this, since. For the Internet Archive's purposes, this dump feature needs to be enabled on ever captured page.

(If I recall our discussion correctly, there was an option that only required adding `r.Proto` and `r.Status` back into colly.Response.)

CorentinB:
@asciimoo I tried to implement individual methods for dumping requests and response, but while it was working for the response, the request was way more complicated because when I was trying to dump it it was already consumed for the execution, so still in the idea of not having a higher memory usage for people who do not care about dumps, I chose to implement a collector setting, that, when set to `true` dump the responses and requests. 
I also implemented basic tests for it, I hope it is enough. I am sadly not very used in Go testing!

Let me know if anything needs to be changed. :)

asciimoo:
I've quickly checked the source code of both `httputil.DumpResponse`[0] and `httputil.DumpRequestOut`[1] and as I see both can be rewritten to be able to provide the same output with the current `colly.Response` and `colly.Request` objects.
As this feature isn't a widely used feature, I'm not sure that modifying colly's internals is the right way to use it.
What about creating an extension[2] or a standalone module with the modified dump functions?

[0] https://golang.org/src/net/http/httputil/dump.go?s=8166:8231#L271
[1] https://golang.org/src/net/http/httputil/dump.go?s=1848:1913#L56
[2] https://github.com/gocolly/colly/tree/master/extensions

CorentinB:
@asciimoo I am trying to do an implementation of `DumpResponse` and `DumpRequestOut` for `colly.Response` and `colly.Request`, but the issue is that it ends up just duplicating a lot of code that already exist in `net/httputil`, and still modifying Colly's internals. For example, to replicate `DumpResponse` behavior in Colly, I need to access the `ContentLenght` and `Status` fields of the `http.Response`.

Another example, to execute a `colly.DumpRequestOut`, we would need to access the `colly.Request` body, that is empty after the execution of the request. The implementation of my idea with the Collector setting few days ago avoid having to expose a `Body []byte` for the Request, like it is done for `Colly.Response`.

So, I am not sure that the extension and DumpResponse/DumpRequestOut re-implementation is the path to go. (I would have preferred to think about it before spending time coding it and even "resetting" the branch for a clean code base, I've been kind of dumb on this one! :smile:)

Let me know what you think. :slightly_smiling_face:

asciimoo:
> I need to access the ContentLength and Status fields of the http.Response

We can expose ContentLength in the Response object and status code is already available as `Response.StatusCode`.

> Another example, to execute a colly.DumpRequestOut, we would need to access the colly.Request body,

Body can be stored before passing it to Colly or you can implement a custom reader which is able to store the content for further readings, like https://golang.org/pkg/io/#TeeReader .

> I would have preferred to think about it before spending time coding

Sorry for the inconvenience, it is a more complex topic than I first thought, my fault.

CorentinB:
> Sorry for the inconvenience, it is a more complex topic than I first thought, my fault.

No problem at all!

> We can expose ContentLength in the Response object and status code is already available as Response.StatusCode.

Yes, but we need `Status`, not `StatusCode`, so is it okay to also expose Status in addition to `ContentLength`?

dvanduzer:
Technically this needs the entire first line of the HTTP response, which I don't think can be accessed ever again: https://golang.org/src/net/http/response.go?s=#L161

Initially I thought this wouldn't be a problem, and that we can reconstruct it if colly.Response has `Status` and `Proto` from http.Response. But I noticed today that net/http _might_ also strip additional whitespace, which means we might fail to capture the bytes exactly as transmitted by the HTTP server. If there is any other way to ensure we are capturing that initial line's pristine bytes, we would like to find it.

StantonMatt:
I checked the current PR head against `origin/master`, and this branch is empty now.

The tip commit is `24969b3` (`Reset branch to master code`), and both GitHub and a local diff show no changed files:

```text
git diff --name-status origin/master...FETCH_HEAD
git diff --stat origin/master...FETCH_HEAD
```

Both commands produced no output. Unless there is a plan to reopen the dump feature on this branch, this PR looks safe to close as stale/empty.

PR #441, opened 2020-01-29
Safely unlock Collector Mutex in defer clause

maciej:
Using defers to unlock mutexes is safer.
If the code holding the mutex ever panics (eg. a bug is introduced)
`defer` will guarantee that it gets unlock and Collector will not
stay in an invalid state.

Since Go 1.14 there is no reason not to use `defer`:
https://twitter.com/janiszt/status/1215601972281253888

Previously performance might have been a concern. It no longer will be

Citing Effective Go:
> Defer

> Go's defer statement schedules a function call (the deferred function) to be run immediately before the function executing the defer returns. It's an unusual but effective way to deal with situations such as resources that must be released regardless of which path a function takes to return. The canonical examples are unlocking a mutex or closing a file.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -724,10 +724,23 @@ func (c *Collector) isDomainAllowed(domain string) bool {
 	return false
 }
 
-func (c *Collector) checkRobots(u *url.URL) error {
+func (c *Collector) robotsMapGet(host string) (robot *robotstxt.RobotsData, ok bool) {
 	c.lock.RLock()
-	robot, ok := c.robotsMap[u.Host]
-	c.lock.RUnlock()
+	defer c.lock.RUnlock()
+
+	robot, ok = c.robotsMap[host]
+	return
+}
+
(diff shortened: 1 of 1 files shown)


colly.go
```diff
@@ -724,10 +724,23 @@ func (c *Collector) isDomainAllowed(domain string) bool {
...
 
-func (c *Collector) checkRobots(u *url.URL) error {
+func (c *Collector) robotsMapGet(host string) (robot *robotstxt.RobotsData, ok bool) {
 	c.lock.RLock()
-	robot, ok := c.robotsMap[u.Host]
-	c.lock.RUnlock()
+	defer c.lock.RUnlock()
+
+	robot, ok = c.robotsMap[host]
+	return
```
rochakgupta, line 732: No need to `return` from the end of the function.

rochakgupta (commented):
Looks good, except for a minor change.

PR #449, opened 2020-02-22
Fix data race in queue

gopherclass:
`Queue.activeThreadCount` is a shared variable. Two goroutines is accessing `Queue.activeThreadCount` without correct synchronization, which can cause data race.


gopherclass:
I opened PR #452. May do I close?

Issue #450, opened 2020-02-22
Infinitely looping queue


gopherclass:
Current queue implementation loops infinitely if queue is empty and active.

https://github.com/gocolly/colly/blob/b1a8ed2f18144f4b70abcfc18a5e58c68a062389/queue/queue.go#L125-L132

![pprof](https://user-images.githubusercontent.com/56159921/75090997-f962ed00-55ab-11ea-8a8e-89c06391dd9b.png)

tobinski:
I think to fix this we need to change the thread goroutine to something like this.
```go
go func() {
	defer wg.Done()
	for r := range q.requestsOut {
                //  this needs to be inside the for loop
		atomic.AddInt32(&q.activeThreadCount, 1)
		r.Do()
		atomic.AddInt32(&q.activeThreadCount, -1)
	}
}()
```

gopherclass:
The following scenario is possible. Queue has one request which takes
longer to complete (e.g. httpbin/delay/10). First, the producer goroutine
sends the request to the consumer goroutine. The consumer executes the
request. Returning to the producer, the producer enters if block since the
queue is empty. q.activeThreadCount is 1. Nested if is not executed and
continue statement is reached. This process is repeated until the request
is completed (for 10 seconds). Your scenario cannot handle the scenario.

My opinion is that we change Storage.GetRequest to be blocking, or use
Queue as Storage when Storage is passed as a argument to queue. I think
either can be a breaking change.



2020년 2월 26일 (수) 오전 2:31, tobinski <notifications@github.com>님이 작성:

gopherclass:
I opened PR #452 to address this issue.

asciimoo:
hopefully fixed by #452

PR #452, opened 2020-02-28
Fix Queue bug

gopherclass:
The existing queue implementation has three drawbacks:
1. Infinite loop (#450)
2. Data race (#449)
3. Last requests may not be executed because of wrong implementation.

The third is a subtle bug. You can see the bug by `go test -v -timeout 10s ./queue` (be sure to checkout master branch and cherry-pick the test commit). The bug might not be shown as the bug is undeterministic.

This pull request addresses the three drawbacks.

I assumes that Storage stored in Queue must not be used by a user while Queue.Run() is running. The assumption can be considered a breaking change to some users. Be careful to merge.

### DIFF
--- a/queue/queue.go
+++ b/queue/queue.go
@@ -3,7 +3,6 @@ package queue
 import (
 	"net/url"
 	"sync"
-	"sync/atomic"
 
 	"github.com/gocolly/colly/v2"
 )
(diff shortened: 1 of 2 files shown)


gopherclass:
CI seems not to test ./queue.

PR #456, opened 2020-03-09
Async delete parameter

zyt312074545:
CollectorOption Async method delete parameter


asciimoo:
You are right, it is an unnecessary parameter, but could you make it optional to avoid api break?

zyt312074545:
You are right, I cannot delete it directly, we need avoid api break, I will update pull request. @asciimoo

PR #460, opened 2020-03-17
Add methods to InMemoryQueueStorage to allow requests without serialization

BenKnigge:
Added the methods AddRequestPointer and GetRequestPointer to allow for working with InMemoryQueStorage without having to use Serialization.

I had to make SerializableRequest public to allow it to be used from the Queue package


BenKnigge:
adding a a bit to this and will create a new pr when finished.

PR #468, opened 2020-04-11
Revisit detection improvement

asalih:
Same request check implemented only GET requests in #77b8884. But  same post bodies also need to be checked and prevent duplicate post requests. Otherwise there is no way to prevent POST requests.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -16,7 +16,6 @@
 package colly
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"crypto/rand"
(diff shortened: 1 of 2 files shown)


asciimoo:
Good improvement, thanks. Could you add tests?

asalih:
You are welcome. I've added post revisit check test func. Also, IsVisited func checks request url only. how about implementing a new func like accepting request url and post body, wdyt?

asciimoo:
> how about implementing a new func like accepting request url and post body, wdyt?

Good idea, but it would be good to avoid api break. Perhaps it could be an optional second argument.

asalih:
> > how about implementing a new func like accepting request url and post body, wdyt?
> 
> Good idea, but it would be good to avoid api break. Perhaps it could be an optional second argument.

I've added HasPosted func. No API break, imo. Also calling Seek func were needed for the reader after my pull request and I've fixed it. Tests running successfully.

PR #476, opened 2020-04-29
Added new project that uses Colly

deven96:
A scraper for movie sites that bypasses ads to return download links


deven96:
Thanks!. Colly made it possible

PR #479, opened 2020-04-29
Fix relative <base> URL

WGH-:
The base itself can be relative, this is documented[1], and rarely
happens on the websites in the wild.

[1] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1057,7 +1057,7 @@ func (c *Collector) handleOnHTML(resp *Response) error {
 		return err
 	}
 	if href, found := doc.Find("base[href]").Attr("href"); found {
-		resp.Request.baseURL, _ = url.Parse(href)
+		resp.Request.baseURL, _ = resp.Request.URL.Parse(href)
 	}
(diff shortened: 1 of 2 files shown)


WGH-:
For example, as of right now, https://feedly.com/i/welcome looks like the following (only relevant parts are shown):

```html
<base href=//s5.feedly.com/ />
<script type="text/javascript" src="//s5.feedly.com/web/master/vendors.2717d1ce4bd5113de5e7.js" defer>
```

Since baseURL is initialized just from `href` value, schema is lost. If you try to resolve the address for the script (i.e. `req.AbsoluteURL("/s/5.feedly.com/web/...")`), you will also get a schemaless URL, where it should have been https instead.

Of course, the problem is not limited to losing schema: https://marketingplatform.google.com/about/enterprise/
```html
<base href="/about/">
<link href="static/css/index.min.css" rel="stylesheet">
```

Here hostname part is lost as well: the resolved URL is just `/about/static/css/index.min.css`.

Issue #482, opened 2020-05-04
Race condition when using Parrellelism and Proxies


i25959341:
Got races when enabling parrellel thread,

Need to use a thread safe array

Previous write at 0x00c000518298 by goroutine 62:
  sync/atomic.AddInt32()
      /usr/local/Cellar/go/1.14/libexec/src/runtime/race_amd64.s:269 +0xb
  github.com/gocolly/colly/v2/proxy.(*roundRobinSwitcher).GetProxy()
      /Users/karlsonlee/go/pkg/mod/github.com/gocolly/colly/v2@v2.0.1/proxy/proxy.go:33 +0x118

asciimoo:
Fixed in #483

PR #483, opened 2020-05-04
Proxy accessing array in a not thread safe way

i25959341:


i25959341:
#482

WGH-:
`AddUint32` returns the previous value of the variable. Can't this be safely rewritten as

```go
index := atomic.AddUint32(&r.index, 1) - 1
u := r.proxyURL[index%uint32(len(r.proxyURLs))]
```
instead?

i25959341:
that is better

WGH-:
Oops, correcting myself:

> AddUint32 returns the previous value of the variable.

It actually returns the new value. The code was (and still is) correct, `-1` is indeed necessary so it starts from the first element, it's just my comment was wrong and misleading.

i25959341:
Ya, I sort of got that when I was correcting it, but ya around 5-6 lines better than before

PR #484, opened 2020-05-07
return current storage.Storage

jony4:


asciimoo:
What is the purpose of this change?

jony4:
> What is the purpose of this change?

when I want to read cookies which had stored in **storage.Storage**, The only way is

`
c.store.Cookies(&url.URL{})
`

but as you see, the `store` cannot be read by others, so I add this method.

asciimoo:
@jony4 you have to pass the storage to the collector when you initialize it. At this point you have access to the storage, so you can use it later without getting it from the collector. I don't find this addition very useful at first, but I'm keeping it open, if others need it too, I'm going to merge it.

PR #490, opened 2020-05-15
Fixed binary files (images, videos and etc.) encoding as a text/plain

melehin:
Bug: encoding a binary file data as a text/plain if a Content-Type charset not empty and equals binary or other

Two different files (curl'ed for debug):
image
image

Two files downloaded by the gocolly.
First red rectangle is a good file, second is a bad file. 
**Look around a "JFIFHH" magic symbol!**
image

### DIFF
--- a/response.go
+++ b/response.go
@@ -76,7 +76,7 @@ func (r *Response) fixCharset(detectCharset bool, defaultEncoding string) error 
 
 		return nil
 	}
-	
+
 	if defaultEncoding != "" {
 		tmpBody, err := encodeBytes(r.Body, "text/plain; charset="+defaultEncoding)
 		if err != nil {
--- a/response.go
+++ b/response.go
@@ -65,14 +65,7 @@ func (r *Response) fixCharset(detectCharset bool, defaultEncoding string) error 
 	if len(r.Body) == 0 {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


WGH- (changes requested):
I agree that this's most likely a bug, but this needs tests.

StantonMatt (commented):
I checked this against current `master` via the synthetic merge ref (`012ee43`). The bug still reproduces on current `master` when a binary response has an image content type and the request/default encoding is set:

```text
Content-Type: image/jpeg; charset=iso-8859-1
default encoding: iso-8859-1
```

A temporary local regression around `Response.fixCharset(false, "iso-8859-1")` changes these bytes on `master`:

```text
want: ffd8ffe04a4649460080
got:  c3bfc398c3bfc3a04a46494600e282ac
```

The same test passes on the merge ref because the binary MIME check now runs before the default-encoding conversion.

Validation on the merge ref:

```text
go test . -run TestPR490BinaryResponseIgnoresDefaultEncoding -count=1 -v
go test ./... -count=1
```

Those passed locally. GitHub reports no hosted checks for this PR.

So the code direction still looks valid on current `master`; I agree with the earlier review that the missing piece is a committed regression test for the binary MIME + default encoding case before merge.

PR #512, opened 2020-06-15
Make allowances for responses with missing ContentLength

### LINKED ISSUE #511: Unnecessary failure when missing server sends broken content-length
If a server is broken and returns an invalid or missing content-length, colly reports "unexpected EOF", despite having received data.

Repro: 

1. use colly to visit https://regexr.com/39p0t
2. res.ContentLength is reported as -1
3. EOF error is returned from httpBackend.Do, even though body has all the content.

(This is an ErrUnexpectedEOF, not a regular EOF, which is why ioutil.ReadAll returns it.)

I would expect colly to return the Response in this case.

majelbstoat:
This makes colly more permissive to servers that don't send a ContentLength header.

Fixes https://github.com/gocolly/colly/issues/511

### DIFF
--- a/http_backend.go
+++ b/http_backend.go
@@ -208,7 +208,10 @@ func (h *httpBackend) Do(request *http.Request, bodySize int, checkHeadersFunc c
 	}
 	body, err := ioutil.ReadAll(bodyReader)
 	if err != nil {
-		return nil, err
+		// Make allowances for broken servers that don't sent a content length.
+		if err != io.ErrUnexpectedEOF || res.ContentLength != -1 {
+			return nil, err
+		}
 	}
 	return &Response{
 		StatusCode: res.StatusCode,


StantonMatt (commented):
I checked this against current `master` via the GitHub merge ref (`59da31d`). It still applies cleanly and does change the #511 behavior: a local-only `httpBackend.Do` repro where the response body returns partial bytes plus `io.ErrUnexpectedEOF` with `ContentLength == -1` fails on current `master` with `unexpected EOF`, but passes on this merge ref and returns the partial body.

Validation on the merge ref:

```text
go test . -run TestPR512UnexpectedEOFWithUnknownLength -count=1 -v
go test ./... -count=1
```

Those passed locally. GitHub reports no hosted checks for this PR.

One merge consideration: the issue discussion narrowed this to broken/truncated gzip responses, while this branch suppresses any `io.ErrUnexpectedEOF` when `ContentLength == -1`. That may be the intended curl/browser-tolerant behavior, but I would make the scope explicit with a committed regression test before merge.

majelbstoat:
I admit to having lost the thread of this after so many years! I forked and applied this patch and have been happily living with it ever since. It may or may not be a “correct” fix, but it got me what I needed. Anyone is welcome to take it on and do whatever they like with it 😊

(Props on a stable codebase where a 6 year old patch still applies cleanly!)

PR #526, opened 2020-07-30
adding docker-slim to the list of open source projects using Colly

kcq:
let me know if you need any other information about the project or if you need more details about the Colly package use


kcq:
> Great project, thanks!

Thanks for gocolly :-)

PR #503, opened 2020-06-06
add chromedp backend

mazhigali:
add chromedp to backend
https://github.com/gocolly/colly/pull/148 
https://github.com/gocolly/colly/issues/4

PR #534, opened 2020-08-28
Defines an HTTPBackend interface to use with Colly.

riverbeecat:
This interface is designed to support using different backends for fetching data, specifically to support a future PR using chromedp as a backend to support parsing sites that rely on .js to populate data.

It's based off of work by @mazhigali in https://github.com/gocolly/colly/pull/503/files, with some simplifications and tweaks to naming conventions. I'm looking to use a chromedp based scraper in a project I'm doing, so wanted to progress that PR given it's gone a little quiet, but happy to work back on it in a different way if you or @mazhigali would prefer.

The intent is to provide a fast follow with an implementation of a chromedp based backend as well as modifications to colly.go to support setting up with different backends, with this first PR setting up for that with a smaller changeset that's hopefully easier to review.

I haven't added tests for this as it's largely just getters/setters rather than substantive code, but I can work back on that as preferred.

Testing was done by running some of the colly examples as well as using the pre-written tests.

Thanks!

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -122,7 +122,7 @@ type Collector struct {
 	scrapedCallbacks         []ScrapedCallback
 	requestCount             uint32
 	responseCount            uint32
-	backend                  *httpBackend
+	backend                  HTTPBackend
 	wg                       *sync.WaitGroup
 	lock                     *sync.RWMutex
 }
@@ -217,7 +217,7 @@ var envMap = map[string]func(*Collector, string){
 		c.DetectCharset = isYesString(val)
 	},
 	"DISABLE_COOKIES": func(c *Collector, _ string) {
-		c.backend.Client.Jar = nil
+		c.backend.SetJar(nil)
 	},
 	"DISALLOWED_DOMAINS": func(c *Collector, val string) {
 		c.DisallowedDomains = strings.Split(val, ",")
@@ -402,10 +402,10 @@ func (c *Collector) Init() {
 	c.store = &storage.InMemoryStorage{}
 	c.store.Init()
 	c.MaxBodySize = 10 * 1024 * 1024
-	c.backend = &httpBackend{}
+	c.backend = &httpClientBackend{}
 	jar, _ := cookiejar.New(nil)
 	c.backend.Init(jar)
-	c.backend.Client.CheckRedirect = c.checkRedirectFunc()
+	c.backend.SetCheckRedirect(c.checkRedirectFunc())
 	c.wg = &sync.WaitGroup{}
 	c.lock = &sync.RWMutex{}
 	c.robotsMap = make(map[string]*robotstxt.RobotsData)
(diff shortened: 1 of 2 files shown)


asciimoo:
Great refactor, thanks! Just a few questions:
 - how many of the http backend interface functions cannot be implemented with chromedp? (E.g. I'm pretty sure that `SetClient` will not work with chromedp, probably `SetJar`/`GetJar` neither)
 - how should we handle the differences of the different backends? Besides chromedp cannot implement some of the backend interfaces function, there are also methods that are only supported by chromedp like `Click`.
 - how can we use our callbacks with chromedp?

riverbeecat:
My intention is to implement a slim version of the backend, basically to parity with the httpClient, so not to support any of the more complex functionality like Click etc. but just to have it provide a headless browser alternative for fetching HTML. There are some areas that won't work precisely (`setClient` is the most notable), but I _think_ we can get something reasonably good.

Let me clean up the draft I have into an actual PR and send that over too, so you can see what I'm thinking.

asciimoo:
There are quite a few differences between the two backends. Wouldn't it be better to implement the headless version as an independent module/package?

riverbeecat:
Happy to do that too, the rationale was to leverage the work that's already done here around things like rate limits, max depth of crawls etc. I can just borrow the patterns and implement a separate lib though.

I'll close this PR and move to that approach 👍

asciimoo:
> I'll close this PR and move to that approach 👍

I'm not saying that this approach isn't working. I just want to be sure that we pick the best solution, that's why I'm asking so many questions. =)
It would be good to collect the similarities and differences of the two backends to be able to make a good decision. A few minor differences are not a dealbreaker, but if we can't use our callback based workflow with the headless backend or if we cannot use crucial elements of the headless browser, like `Click`, then I think handling the two backends similarly isn't the best way.

Also, this PR can be still relevant even if we decide to go with a different model for the headless browser, because it makes us able to use other "static" http backends, like fasthttp. So I wouldn't close this PR.

riverbeecat:
Sure, I'm happy to re-open this and implement this change. Do you have any feedback re. the PR in terms of just providing an HTTPBackend interface?

It's more work than makes sense for me to do to take on a thoroughgoing analysis of how to integrate chromedp with colly and allow things like click / select etc. to work though. I'll handle my use case for that w/ a fork or a parallel lib.

asciimoo:
> Sure, I'm happy to re-open this and implement this change.

Great, thanks!

> Do you have any feedback re. the PR in terms of just providing an HTTPBackend interface?

It looks good to me at first glance, but I have to do a deeper review before I approve it.

> It's more work than makes sense for me to do to take on a thoroughgoing analysis of how to integrate chromedp with colly and allow things like click / select etc. to work though

It's understandable, this is a big topic indeed.

>  I'll handle my use case for that w/ a fork or a parallel lib.

Can I ask what is your use case that requires chromedp as a backend without its functionality?

riverbeecat:
I want to parse content from a single page application that uses a complex set of API calls to render the DOM. It's possible to parse the API calls, but that would require a fair bit of analysis and it seems easier to write something that allows me to work with the DOM and that I can potentially extend in the future to support interaction with the page as needed.

asciimoo:
Why don't you use chromedp directly for this task?

riverbeecat:
Because the task is basically a scraping task - walking a website and extracting content - and having access to tools like rate limiting across domains and other things scrapers provide is helpful.

Practically, that's what I'm intending to do though - it's what I meant by implementing my own lib - I'll just write a minimal version with the couple of patterns (rate limits etc.) from scrapers that make sense for my use case.

asciimoo:
I see. Thanks for your answer. I'm afraid that colly's rate limiting would not work with the headless browsers internal network requests by default, so probably it's easier to handle rate limits with your own lib.

PR #537, opened 2020-08-30
update coursera example

ysung6:
Could you consider updating this example?
Seems like Coursera updated their website. This example was returning an empty list as the final result.

Changes:

- Update querySelectors
- save output to json file


asciimoo (approved):
Great, thanks!

PR #548, opened 2020-10-09
Add context.Context support

WGH-:
This allows one to pass context.Context to colly.Collector, which in turn will passed to stdlib's http.Request.

This enables colly scraping to be cancelled cleanly at any time.

See #255 #240


WGH-:
Given that the word `Context` collides with already existing type in `colly`, I couldn't think of anything better than calling the `CollectorOption` function `StdlibContext`. Better ideas are more than welcome.

asciimoo (approved):
Great, thanks! I think `StdlibContext` is perfectly fine

PR #561, opened 2020-11-24
Fix handling of redirected URLs in OnResponseHeaders

WGH-:
When `OnResponse` callback is called, the request URL is updated to reflect the redirect that has been followed (if any).

Do the same for `OnResponseHeaders`, for consistency.


asciimoo (approved):
Good fix, thanks!

PR #562, opened 2020-11-26
Fix excessive memory usage

x-wesley:
When the website has too much text content, the response will take up too much memory. 

Use sync.Pool to reduce memory applications.

Thanks.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -665,8 +665,20 @@ func (c *Collector) fetch(u, method string, depth int, requestData io.Reader, ct
 		return !request.abort
 	}
 
+	var response *Response
+	var err error
+	defer func() {
+		if response != nil {
+			response.Body = []byte{}
+			response.Headers = nil
+			response.Request = nil
+			response.Ctx = nil
+			response.Trace = nil
+			responsePool.Put(response)
+		}
+	}()
 	origURL := req.URL
-	response, err := c.backend.Cache(req, c.MaxBodySize, checkHeadersFunc, c.CacheDir)
+	response, err = c.backend.Cache(req, c.MaxBodySize, checkHeadersFunc, c.CacheDir)
 	if proxyURL, ok := req.Context().Value(ProxyURLKey).(string); ok {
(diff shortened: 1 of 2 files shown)


WGH-:
Does it really help that much, given that the `[]byte` body, headers, and even `net/http.Response` objects are still allocated?

x-wesley:
> Does it really help that much, given that the `[]byte` body, headers, and even `net/http.Response` objects are still allocated?

Yeah, when there are a lot of URLs in a website and a lot of content on the webpage, a lot of memory space will be requested in a short time.

asciimoo:
@x-wesley do you have measurements about how much improvement this change causes?

x-wesley:
> @x-wesley do you have measurements about how much improvement this change causes?

yeah
In May, we have a scheduled task. We will receive a high memory usage alarm (>80%) every time the task is executed.
The problem is located through pprof, and the peak memory usage after modification is about 50%-60%.

StantonMatt (commented):
I checked this branch locally. The existing suite still passes (`go test ./... -count=1`) and `git diff --check origin/master...HEAD` is clean, but the pooling change introduces a response-lifetime regression.

The branch returns the `*Response` to `responsePool` in `fetch`'s deferred cleanup and clears `Body`, `Headers`, `Request`, `Ctx`, and `Trace`. A local probe that saves the `*Response` from `OnResponse` and reads it after `Visit` returns passes on current `origin/master`; on this branch it fails immediately because the saved response has already lost `Request` after `Visit` returns, with `Body`/`Headers` also cleared by the same defer.

Since `OnResponse` hands user code a `*Response` and there is no ownership boundary saying it is only valid during the callback, this would break callers that keep the response or inspect it after the visit completes. I think the memory-reuse strategy needs a different ownership model before the memory measurements matter.

Two smaller current-state notes: the branch now conflicts with current master in `colly.go` and `http_backend.go`, and `gofmt -l colly.go http_backend.go` reports both touched files.

Issue #339, opened 2019-05-28
GetProxy in proxy.go is not called on every request


dabba39:
, DisableKeepAlives in Transport solves this problem. But there should be a more general solution i think?

makelove:
same problem

I think it is a Big BUG of colly
unable to use proxy proxy freely

asciimoo:
Could you provide code which reproduces the issue?

IOsetting:
> Could you provide code which reproduces the issue?

I got this problem too, you can reproduce this issue exactly with the example http://go-colly.org/docs/examples/proxy_switcher/
GetProxy will be called only once so each time the same proxy will be used.

superiss:
I got the same issue. with the rotating proxy. http://go-colly.org/docs/examples/proxy_switcher/

this for now helped.
https://github.com/gocolly/colly/issues/343#issuecomment-569526916

wowinter13:
@asciimoo I think this issue could be closed

asciimoo:
Thanks!

PR #567, opened 2020-12-17
SetProxyFunc: fix proxies rotation

wowinter13:
Hey!
I think `SetProxyFunc()` could simply override `DisableKeepAlives` and set it `true` as the default value. Otherwise, it will rotate proxies only on failed requests.

Also, you can reproduce this issue exactly with the example http://go-colly.org/docs/examples/proxy_switcher/

This PR addresses the issue #339


wowinter13:
@asciimoo What do you think?

asciimoo (approved):
Great, thanks!

PR #568, opened 2020-12-24
export RequestCount/ResponseCount

wsw0108:


wsw0108:
can be count in OnRequest/OnResponse callbacks.

PR #574, opened 2021-01-13
Add httpBackend cache response callback

slashapp:
The checkHeadersFunc is called before response cached body, it can check cache expire or force cache invalidation.

```
c.OnResponseHeaders(func(r *colly.Response) {
	respTime, err := time.Parse(http.TimeFormat, r.Headers.Get("Date"))
	if err != nil || time.Now().UTC().Sub(respTime) > maxAge {
		r.StatusCode = 510 // invalidate the cache
	}
})
```


asciimoo (approved):
Thanks!

PR #576, opened 2021-01-20
Update openedx_courses.go

BigManing:
Update some goquerySelector  .   Because the target website HTML has been updated .

such as :

image

image


asciimoo (approved):
Thanks!

Issue #343, opened 2019-06-03
Proxy Authentication


andystroz:
I was wondering how to use the proxy functionality with Basic Auth? I've modified the callback function in insert the base64 encoded username and password with the `Proxy-Authorization` header. I have also verified the same HTTP request works when using `https://httpbin.org/ip` in cURL but Colly returns `Error: Proxy Authentication Required`.

andystroz:
Just saw [this issue](https://github.com/gocolly/colly/issues/338) and I think its related because `http://httpbin.org/ip` works but `https://httpbin.org/ip` does not. The website I am looking to scrape uses `https`

suisun2015:
Hi Andrew
I met same error. Did you figure it out?

andystroz:
Yea! I did end up getting it working, hopefully this still helps.
```
c := colly.NewCollector()
c.WithTransport(s.HTTPTransport)
```

Where `s.HTTPTransport` is:
```
&http.Transport{
	Proxy: rp,
	DialContext: (&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
		DualStack: true,
	}).DialContext,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
	TLSNextProto:          nil,
}
```

Where `rp` is a [function that specifies a proxy to use](https://golang.org/src/net/http/transport.go).
```
// AuthenticatedRoundRobinProxyHTTP returns the proxy fuction used by http for an authenticated round robin proxy
func AuthenticatedRoundRobinProxyHTTP(proxyURLs []string, username string, password string) (func(*http.Request) (*url.URL, error), error) {
	roundRobinSwitcher, err := collyProxy.RoundRobinProxySwitcher(proxyURLs...)
	if err != nil {
		return nil, err
	}
	return (&authenticatedRoundRobinSwitcher{roundRobinSwitcher, username, password}).GetAuthenticatedProxy, nil
}
```

Where `proxyURLs []string` is a list of socks5 proxy URLs ex. `socks5://127.0.0.1:1337,socks5://127.0.0.1:1338` and `&authenticatedRoundRobinSwitcher{roundRobinSwitcher, username, password}).GetAuthenticatedProxy` returns `(request *http.Request)` returns the request with the correct headers and the URL returned from the colly round robin switcher.

```
[...]

PR #578, opened 2021-01-22
Add a AuthenticatedRoundRobinProxyHTTP method to authenticate proxies

wowinter13:
This PR addresses the issue #343 

Hey! 
Right now colly doesn't have an ability to authenticate proxies out of the box. So I added this functionality:)
Also I decided not to modify `proxy_switcher.go` because it would break an example without having correct credentials.
(Also my deepest gratitude to @andystroz)

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -76,12 +76,12 @@ func RoundRobinProxySwitcher(ProxyURLs ...string) (colly.ProxyFunc, error) {
 	return (&roundRobinSwitcher{urls, 0}).GetProxy, nil
 }
 
-// AuthenticatedRoundRobinProxyHTTP decorates RoundRobinProxySwitcher and sets proxy credentials.
+// AuthenticatedRoundRobinProxySwitcher decorates RoundRobinProxySwitcher and sets proxy credentials.
 // See RoundRobinProxySwitcher for more details on rotating proxies.
 // This method sets correct "Proxy-Connection" and "Proxy-Authorization" headers.
 // "Proxy-Authorization" contains the credentials (username, password) to authenticate
 // a user agent to a proxy server.
-func AuthenticatedRoundRobinProxyHTTP(username string, password string, ProxyURLs ...string) (func(*http.Request) (*url.URL, error), error) {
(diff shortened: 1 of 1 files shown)


wowinter13:
@asciimoo Also we could rewrite `RoundRobinProxySwitcher` and make credentials to be nullable. Therefore we would be able to delete `AuthenticatedRoundRobinProxySwitcher`. (but imho for now that is better to have both methods)

WGH-:
I personally don't see why it has to be built in into Colly. You already can install a custom HTTP RoundTripper with any behaviour you want.

wowinter13:
Hey @WGH- !
Most crawlers have the functionality to authenticate proxies. At least, Scrapy.
The situation is a little bit inconvenient when you have a built-in method to rotate proxies and don't have any methods to do basic auth for these proxies.

Also if you check the issue #343, you will see some demand for this feature.

wowinter13:
@asciimoo ?

samirsss:
Any thoughts on this approach - i'd like this to be included as part of Gocolly too.

StantonMatt (commented):
I rechecked this because #343 still comes up and the branch is old enough that the current behavior is easy to miss.

The PR head `710cd77` still builds locally:

```text
go test ./proxy -count=1 -v
go test ./... -count=1
```

I also tested a temporary merge with current `origin/master`; the merge was clean and the same `go test ./proxy -count=1 -v` / `go test ./... -count=1` checks passed.

The part I would resolve before adding a new API is whether this helper is still needed. Current `RoundRobinProxySwitcher` already accepts credentials in the proxy URL, and `net/http` sends the proxy auth from that URL. I verified this with a local fake proxy using these inputs:

```text
proxy URL: http://user:pass@127.0.0.1:<port>
HTTP target: http://example.invalid/
HTTPS target: https://example.invalid/
```

Both visits sent this header to the proxy, including on the HTTPS `CONNECT` request:

```text
Proxy-Authorization: Basic dXNlcjpwYXNz
```

So for the original #343 use case, a credentialed proxy URL such as `http://user:pass@host:port` appears to cover authenticated HTTP proxies without a separate wrapper. If maintainers still want a convenience helper for sharing one username/password across multiple proxy URLs, I think this PR would need tests and docs for that narrower API rather than only adding the function.

Issue #581, opened 2021-01-28
Out-of-date version of xmlquery introduces security vulnerability


lamados:
This package currently uses `xmlquery` v1.2.4 which is affected by a high-severity vulnerability (as described [here](https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMANTCHFXXMLQUERY-608944)). This was talked about in [this issue](https://github.com/antchfx/xmlquery/issues/39) and is no longer present in v1.3.1 and above, so `colly` should probably be updated to use the latest version (v1.3.3).

asciimoo:
Fixed in #582, thanks!

PR #582, opened 2021-01-29
Updated xmlquery to close a known security vulnerability

lamados:
As described in #581 xmlquery versions before v1.3.1 are vulnerable to denial-of-service attacks, so to mitigate that I've updated xmlquery to the latest version (v1.3.3).


asciimoo (approved):
Great, thanks!

PR #588, opened 2021-02-14
Change the type assersion for Attr 

moritamori:
Attr was changed from xml.Attr to xmlquery.Attr in [version 1.3.4](https://github.com/antchfx/xmlquery/commit/c52bb6c871ec4eead23d6f1ad4dd27e7b77f6514), so I changed the type assertions in xmlelement.go.

And also added test for the following codes.
https://github.com/gocolly/colly/blob/master/colly.go#L1140-L1157


asciimoo (approved):
Great, thanks!

moritamori:
conflict resolved

PR #589, opened 2021-02-16
Remove tabs and newlines from URLs

WGH-:
This might sound weird, but both URL standard[1] specifies it,
and browsers do that as well.

Although the standard specifies it as a "validation error",
this is not a hard error.

This actually happens in the wild: as of now, this Google's page[2]
has the following fragment:

    <a class="glue-header__link"
                                  href="/intl/ru_ALL
    /drive/download/"
    >

Yes, the newline here is in the middle of the link, and browsers
do ignore it.

[1] https://url.spec.whatwg.org/#concept-basic-url-parser
[2] https://www.google.com/intl/ru/drive/download/

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -549,7 +549,7 @@ func (c *Collector) UnmarshalRequest(r []byte) (*Request, error) {
 }
 
 func (c *Collector) scrape(u, method string, depth int, requestData io.Reader, ctx *Context, hdr http.Header, checkRevisit bool) error {
-	parsedURL, err := url.Parse(u)
+	parsedURL, err := url.Parse(RemoveAsciiTabAndNewlines(u))
 	if err != nil {
 		return err
 	}
@@ -1074,7 +1074,7 @@ func (c *Collector) handleOnHTML(resp *Response) error {
 		return err
 	}
 	if href, found := doc.Find("base[href]").Attr("href"); found {
-		baseURL, err := resp.Request.URL.Parse(href)
+		baseURL, err := resp.Request.URL.Parse(RemoveAsciiTabAndNewlines(href))
 		if err == nil {
 			resp.Request.baseURL = baseURL
 		}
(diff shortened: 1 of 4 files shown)


WGH-:
I know that this solution looks rather dirty, so it probably would be better to not merge it before discussing it a bit.

asciimoo:
I don't think that it is dirty, it looks good to me. The only addition that we might consider is making an option to allow users to disable this feature. I don't know if it makes sense, but perhaps someone wants these white spaces in the URL for some reason.

rochakgupta:
One observation I can make is that the logic to parse the URL is in `func Parse(rawurl string) (*URL, error)`. So, perhaps it would be better to strip tabs and newlines in it. Feel free to correct me if I'm wrong.

WGH-:
> Also, since the created function isn't called anywhere else, wouldn't it be better to make it private?

Yes. I initially left it exported because I wanted to use it in my code as well: `Request.AbsoluteURL` doesn't report errors (which sadly can't be fixed without breaking API), so I had to normalize and resolve relative URLs myself.

In any case, I agree that it's better not add more exported functions without good reason. I can live with this function copy-pasted until v3 comes and fixes the API issues that causes me to require the function to begin with :).

WGH-:
> The only addition that we might consider is making an option to allow users to disable this feature. I don't know if it makes sense, but perhaps someone wants these white spaces in the URL for some reason.

I believe they should never appear in URL in unescaped forms, and this code doesn't touch their escaped forms. In theory someone might have use cases insane enough to rely on that some servers do accept unescaped tabs AND might treat them differently than escaped ones, but I'd argue that in this scenario he better use custom code instead of generic crawler framework.

asciimoo:
I understand, that servers don't accept these urls, but what if someone wants to create a checker that displays malformed urls?

WGH-:
> I understand, that servers don't accept these urls, but what if someone wants to create a checker that displays malformed urls?

This PR doesn't touch HTML attributes, so it doesn't prevent one from inspecting ther original values. Tabs and newlines are stripped only when calling `Visit` and related methods. (if I get your question right)

asciimoo:
My bad, you are right. Please fix the golint issues and it's ready to merge

asciimoo:
@WGH- could you rebase this PR?

WGH-:
Sorry, I forgot to update the status of this PR. With #673, this PR is no longer necessary.

PR #592, opened 2021-03-10
Fix default User-Agent when using custom headers

WGH-:
When using non-nil `hdr` parameter with `c.Request` method, and not setting User-Agent header explicitly, it would be left unset (when it should've been set to `c.UserAgent`). This means that Go HTTP client would use the default "Go-http-client/2.0" string.

Suppressing the header altogether is still possible by setting it to empty string.


asciimoo (approved):
Good point, thanks!

PR #605, opened 2021-04-17
Update User-Agent strings for Desktop

guessi:
- Removed Firefox version earlier than 2020
- Removed Google Chrome version earlier than 2020
- Added support for Microsoft Edge
- Updated Mac OS version strings (10.13.x ~ 12.2.x)
- Extended support for Android 10/11
- Fixed incorrect OS version string for Windows 10

### DIFF
--- a/extensions/random_user_agent.go
+++ b/extensions/random_user_agent.go
@@ -3,6 +3,7 @@ package extensions
 import (
 	"fmt"
 	"math/rand"
+	"strings"
 
 	"github.com/gocolly/colly/v2"
 )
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Great, thanks!

samirsss:
@guessi  I was wondering if you could update the user-agent strings again please - https://github.com/gocolly/colly/issues/762

When scraping websites - i am getting htmls - where its saying browser isnt supported. Ideally we should keep newer items and get rid of older ones. thanks!

guessi:
@samirsss #765 should address the issue you mentioned.

PR #618, opened 2021-06-04
new. add reset callback function

pashifika:
Reset function needed under certain circumstances


StantonMatt (commented):
I checked the current head `f46420f` locally.

The submitted branch still passes the existing suite:

```text
go test ./... -count=1
```

but I see a few blockers before this is mergeable.

First, it conflicts with current `master` in `colly.go`. That matters here because current `master` has newer callback handling around cloned callback slices and detach helpers, so this cannot be evaluated as a clean reset-only patch anymore.

Second, the branch changes the public `OnHTML` API from:

```go
OnHTML(selector string, func(*HTMLElement))
```

to:

```go
OnHTML(selector, uniqueKey string, func(string, *HTMLElement))
```

A local compile probe using the existing two-argument form fails with:

```text
not enough arguments in call to c.OnHTML
have (string, func(e *HTMLElement))
want (string, string, HTMLCallback)
```

That is a breaking API change, and it looks separate from the reset-callback feature.

Third, the reset methods themselves do not appear to have committed tests (`rg 'Reset.*Call|ResetHtml|ResetRequest|ResetResponse|ResetXML'` only finds `reset.go`). `gofmt -l` also reports the touched Go files, including `reset.go`, `colly.go`, `colly_test.go`, `cmd/colly/colly.go`, `debug/webdebugger.go`, `queue/queue.go`, and `unmarshal.go`.

I think this would need to be rebased and reduced to the reset API, with tests for the reset behavior, before the public API/keying idea is considered separately.

PR #619, opened 2021-06-05
Bypass request delay when request is cancelled

waltton:
Before this PR: after cancelling a context for a collector all requests failed with "context canceled" but the delay set by the limit rules still applies.
The intention here is to terminate the collector as soon as possible after its context finishes.

### DIFF
--- a/http_backend.go
+++ b/http_backend.go
@@ -170,7 +170,12 @@ func (h *httpBackend) Cache(request *http.Request, bodySize int, checkHeadersFun
 func (h *httpBackend) Do(request *http.Request, bodySize int, checkHeadersFunc checkHeadersFunc) (*Response, error) {
 	r := h.GetMatchingRule(request.URL.Host)
 	if r != nil {
-		r.waitChan <- true
+		select {
+		case r.waitChan <- true:
+		case <-request.Context().Done():
+			return nil, request.Context().Err()
+		}
+
 		defer func(r *LimitRule) {
 			randomDelay := time.Duration(0)
 			if r.RandomDelay != 0 {
--- a/http_backend.go
+++ b/http_backend.go
@@ -18,6 +18,7 @@ import (
 	"crypto/sha1"
 	"encoding/gob"
 	"encoding/hex"
+	"errors"
 	"io"
 	"io/ioutil"
 	"math/rand"
@@ -173,7 +174,7 @@ func (h *httpBackend) Do(request *http.Request, bodySize int, checkHeadersFunc c
 		select {
 		case r.waitChan <- true:
 		case <-request.Context().Done():
-			return nil, request.Context().Err()
+			return nil, errors.New("context canceled; early return")
 		}
 
 		defer func(r *LimitRule) {
--- a/http_backend_test.go
+++ b/http_backend_test.go
@@ -15,6 +15,13 @@ import (
 )
 
 func TestHTTPBackendDoCancelation(t *testing.T) {
+	rand.Seed(time.Now().Unix())
+
+	// rand up to 10 to not extend the test duration too much
+	p := 1 + rand.Intn(5)        // p: parallel requests
+	n := p + p*rand.Intn(10)     // n: after n, cancel will be called; ensure 1 calls per worker + rand
+	c := n + p*2 + rand.Intn(10) // c: total number of calls; ensure 2 calls per worker after cancel is called + rand
+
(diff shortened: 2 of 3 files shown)


http_backend.go
```diff
@@ -170,7 +171,12 @@ func (h *httpBackend) Cache(request *http.Request, bodySize int, checkHeadersFun
 func (h *httpBackend) Do(request *http.Request, bodySize int, checkHeadersFunc checkHeadersFunc) (*Response, error) {
 	r := h.GetMatchingRule(request.URL.Host)
 	if r != nil {
-		r.waitChan <- true
+		select {
+		case r.waitChan <- true:
+		case <-request.Context().Done():
+			return nil, errors.New("context canceled; early return")
```
WGH-, line 177: I believe earlier version of this PR returned `ctx.Err()` instead. Why did you decide to return a new error instead?

http_backend_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestHTTPBackendDoCancelation(t *testing.T) {
+	rand.Seed(time.Now().Unix())
```
WGH-, line 18: It's better to not have any indeterminate behaviour in tests. Either just make make `p`, `n`, and `c` some constants, or, if you're unsure, use `math/rand.New()` with some constant seed and run your test multiple times using the same generator.

waltton: Made a commit with a hardcoded seed. :)

http_backend.go
waltton: Yes, that was the case. Usually i would use wrap from `github.com/pkg/errors` and just add to the error but the library is not in use here.
My first try at writing the code I was asserting against the time that it took to run all the calls and check if they return asap after the cancel; worked fine locally but fail on CI and yea, check like that is really unreliable.
So my second attempt its based on the error message; to check if the error was the normal error from the request or if it was the early return; I had to make the early return message different from `ctx.Err()` cause that is what we get from the http client that tries to run with a cancelled request.

WGH-: Sorry for the late response. I don't think tests are good enough reason to return different errors (depending on where it happened) when context is cancelled. Are you aware of any other library that does *not* return `ctx.Err()` unchanged?

http_backend_test.go
WGH-: I'd still advise you to use `rand.New()`. Less global state - the better.

http_backend.go
WGH-: >  My first try at writing the code I was asserting against the time that it took to run all the calls and check if they return asap after the cancel; worked fine locally but fail on CI and yea, check like that is really unreliable.

I suppose you could make a test like this:
- Make the test HTTP handler sleep before returning the response, maybe a second or more.
- Set low rate limit, say, 1 req/second.
- Spawn lots of goroutines. Doesn't have to be too much, though, maybe 20-30 or so is enough.
- Cancel context.
- Assert that cancellation does happen in 5 seconds. This value seems high enough so the test would be reliable, but low enough so the test won't accidentally pass without the fix.

What do you think?

StantonMatt (commented):
I rechecked this because the cancellation behavior is still useful and the branch is unusually old.

On the PR head `a1e4bc8`, the existing focused test is stable locally:

```text
go test . -run TestHTTPBackendDoCancelation -count=20 -v
go test ./... -count=1
```

Those passed, and GitHub reports no checks for this branch.

I also tested a local merge of this PR onto current `origin/master` (`908384f`). The merge itself was clean, but the new test needs one mechanical rebase update because current `httpBackend.Do` now takes separate request-header and response-header callbacks. As-is, current master + this PR fails to build with:

```text
http_backend_test.go:57:33: not enough arguments in call to backend.Do
```

After changing the test helper from the old `(req, statusCode, header)` callback to:

```go
checkRequestHeadersFunc := func(req *http.Request) bool { return true }
checkResponseHeadersFunc := func(req *http.Request, statusCode int, header http.Header) bool { return true }
_, err := backend.Do(req, 0, checkRequestHeadersFunc, checkResponseHeadersFunc)
```

…the current-master merge passes the same focused `-count=20` run and `go test ./... -count=1`. So the behavior still looks valid; it just needs that test-signature refresh before maintainers can merge it cleanly today.

PR #629, opened 2021-06-29
fix: generated code has compile error

darjun:
OnRequest/OnResponse/OnError these three callbacks don't has element-selectors arg, so remove


WGH- (approved):

WGH-:
Not sure if anyone ever used these templates, but thanks!

PR #638, opened 2021-08-10
moving filterings to one function (checkFilters) and use it in case of redirect too

joelazar:
The url filters and disallowed url filters were not applied to the requests, which were redirected. I fixed it with this pr + added a test for it too. Let me know what you think about it.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1293,7 +1293,7 @@ func (c *Collector) Clone() *Collector {
 func (c *Collector) checkRedirectFunc() func(req *http.Request, via []*http.Request) error {
 	return func(req *http.Request, via []*http.Request) error {
 		if err := c.checkFilters(req.URL.String(), req.URL.Hostname()); err != nil {
-			return err
+			return fmt.Errorf("Not following redirect to %q: %w", req.URL, err)
 		}
 		if c.redirectHandler != nil {
 			return c.redirectHandler(req, via)
--- a/colly.go
+++ b/colly.go
@@ -713,18 +713,8 @@ func (c *Collector) requestCheck(u string, parsedURL *url.URL, method string, re
 	if c.MaxDepth > 0 && c.MaxDepth < depth {
 		return ErrMaxDepth
 	}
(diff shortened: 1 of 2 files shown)


colly.go
```diff
@@ -1285,10 +1292,9 @@ func (c *Collector) Clone() *Collector {
 
 func (c *Collector) checkRedirectFunc() func(req *http.Request, via []*http.Request) error {
 	return func(req *http.Request, via []*http.Request) error {
-		if !c.isDomainAllowed(req.URL.Hostname()) {
-			return fmt.Errorf("Not following redirect to %s because its not in AllowedDomains", req.URL.Host)
+		if err := c.checkFilters(req.URL.String(), req.URL.Hostname()); err != nil {
+			return err
```
WGH-, line 1296: I'd preserve "not following redirect to" part, as it's handy to know that it's redirect that failed the check. E.g. `fmt.Errorf("Not following redirect to %q: %w", req.URL, err)`.

(with `%w`, you can still write the test properly, except you'll need to use `errors.Is`)

colly_test.go
```diff
@@ -101,7 +103,6 @@ func newTestServer() *httptest.Server {
 
 	mux.Handle("/redirect", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		http.Redirect(w, r, "/redirected/", http.StatusSeeOther)
-
```
WGH-, line 104: I agree that all of these formatting changes are good, but let's keep only the relevant changes in the patch.

WGH- (changes requested):

colly.go
joelazar: Thanks, corrected now.

colly_test.go
joelazar: Yep, sorry, I totally forgot about my auto-format settings when I did this pr. I'm using [gofumpt](https://github.com/mvdan/gofumpt) BTW.

WGH- (approved):

PR #655, opened 2021-10-17
Add schedule feature and tests

schmorrison:
Thanks for maintaining this package!

Simple scheduling mechanism to periodically start a `Collector`'s scraping routine. The entire schedule group can be cancelled via `StopSchedules` or cancelling the `Collector.Context`. When a stop is executed, all currently running schedules will finish naturally (not at once). Currently, the `Collector`'s registered callbacks are **_not_** aware of the `scheduleContext` cancellation via `StopSchedules`.

I'm sure you will like me to tweak the implementation here.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -114,6 +114,12 @@ type Collector struct {
 	// to support clean cancellation of scraping.
 	Context context.Context
 
+	// A list of scheduled scraping routines.
+	//
+	// Cron expressions are parsed via package
+	// https://github.com/adhocore/gronx#cron-expression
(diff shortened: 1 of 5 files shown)


sbabaei (approved):

WGH-:
This uses public Colly API only, you could do it from your own package just fine, and I'm not yet convinced this is frequent enough use case to be integrated into the main code base.

amirrmonfared (approved):

amirrmonfared (approved):

PR #675, opened 2022-01-12
Work around invalid URL encoding in path

WGH-:
Go `net/http` cannot send HTTP requests containing invalid URL encoding in path (e.g. bare percent)
at all[1]. Browsers send a bare percent in such scenario, and do not implicitly autoencode it.

Until the upstream issue is resolved somehow, we have only two alternatives: either fail to fetch such URLs, or at least attempt the autoencoded variant. Lots of webservers handle them the same way, so it's worth trying.

There aren't too many websites with invalid URL encoding in path component, though.

[1] https://github.com/golang/go/issues/29808

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -260,6 +260,8 @@ var envMap = map[string]func(*Collector, string){
 	},
 }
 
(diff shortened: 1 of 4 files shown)


asciimoo (approved):
Thanks!

PR #678, opened 2022-01-28
Code cleanup

WGH-:
There are several commits, which are largely independent, but they touch the same code, so I thought it would be better to batch them together into single pull request instead of doing 2 extra rebases. Of course, if some of them are more questionable than others, I can split them to separate PRs. Please see the commit messages themselves for the description of what and why I attempted to do.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"hash/fnv"
 	"io"
-	"io/ioutil"
 	"log"
 	"net/http"
 	"net/http/cookiejar"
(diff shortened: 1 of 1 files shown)


colly.go
```diff
@@ -562,40 +561,28 @@ func (c *Collector) scrape(u, method string, depth int, requestData io.Reader, c
...
-		Header:     hdr,
-		Body:       rc,
-		Host:       host,
+		req.Host = hostHeader
 	}
 	// note: once 1.13 is minimum supported Go version,
 	// replace this with http.NewRequestWithContext
 	req = req.WithContext(c.Context)
-	setRequestBody(req, requestData)
+	if err := c.requestCheck(u, parsedURL, method, req.GetBody, depth, checkRevisit); err != nil {
```
asciimoo, line 583: Probably I'm missing something, but I can't find where the `req.GetBody` function is implemented. Pls help. =]

WGH-: `GetBody` is set by `NewRequest`: https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/net/http/request.go;l=889-930;drc=refs%2Ftags%2Fgo1.17.8

This code fragment used to be copy-paste in Colly (`setRequestBody`) and was removed in my previous commit.

asciimoo (approved):
Great, thanks!

Issue #405, opened 2019-11-22
Redirected Url ignoring AllowURLRevisit


LeMoussel:
despite the fact that `collector.AllowURLRevisit = false` is positioned, after visiting an redirected Url,  it's possible to visit again the target url.
eg :
1) http://f4b1.com/mysql-comment-changer-la-configuration-de-sql-mode-pour-une-session is redirected to http://f4b1.com/sql/mysql-comment-changer-la-configuration-de-sqlmode-pour-une-session
2) http://f4b1.com/sql/mysql-comment-changer-la-configuration-de-sqlmode-pour-une-session is visting
3) It's possible to do `c.Visit('http://f4b1.com/sql/mysql-comment-changer-la-configuration-de-sqlmode-pour-une-session')`. URL isn't rejected

WGH-:
Fixed by #681

PR #681, opened 2022-02-01
Fix redirects ignoring AllowURLRevisit=false 

WGH-:
This commit introduces a breaking change: `ErrAlreadyVisited` is replaced
with `AlreadyVisitedError`, which allows the user to know the redirect
destination, which might not match the URL passed to `Visit` when multiple
redirects are followed.

See #405

Currently draft because it's based on code cleanup made in #678


WGH-:
No longer draft due to #678 being merged.

asciimoo (approved):
Awesome, thanks!

PR #691, opened 2022-03-06
Host & Custom Header Support

### LINKED ISSUE #564: Unable to set the Host header
When I log out `httputils.DumpRequest` in `http_backend.go`, I can see that the Host header was never set from the OnRequest callback.

For our project, we need to set the Host header on each visit, diving into the code, it seems like that's not possible unless I am mistaken. Any solutions or should I submit a pull request?

ErikOwen:
Fixes #564

This PR adds support for passing in custom HTTP headers for the crawler to use, and overriding the `Host` HTTP header. This can be useful when attempting to crawl an IP address while targeting a specific virtual host used in the web server on that IP address.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -56,6 +56,8 @@ type CollectorOption func(*Collector)
 type Collector struct {
 	// UserAgent is the User-Agent string used by HTTP requests
 	UserAgent string
+	// Custom headers for the request
+	Headers *http.Header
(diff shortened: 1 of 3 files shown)


hakluke:
I really would love to see this implemented

asciimoo (approved):
Nice, thanks!

PR #713, opened 2022-07-06
Bump whatwg-url version

WGH-:
Includes fix for quadratic worst case when parsing some URLs, like enormous data URLs.


WGH-:
Safe change, shouldn't require proper code review.

PR #722, opened 2022-08-16
feat: Add MaxRequests paramete

guilhem:
Signed-off-by: Guilhem Lettron <guilhem@barpilot.io>


WGH- (approved):
It's a reasonable thing to have given we already have `MaxDepth`.

PR #723, opened 2022-08-18
Add example `istockphoto`

QIN2DIM:


QIN2DIM (commented):
1

shosukenoma (approved):

StantonMatt (commented):
I checked this against current `master` with a clean no-commit merge. The example package still compiles when the live-site tests are skipped, and the normal repo suite stays green:

```text
go test ./_examples/istockphoto -run '^$' -count=1 -v
go test ./... -count=1
git diff --check origin/master
```

Those passed locally. I did not run the submitted `TestStandardDownloader`/similar tests as-is because they scrape `istockphoto.com` and write cache files, so they are integration tests rather than CI-safe unit tests.

I see two concrete blockers before this would be merge-ready:

```text
gofmt -l _examples/istockphoto/downloader/downloader.go _examples/istockphoto/downloader/memory.go _examples/istockphoto/downloader/typing.go _examples/istockphoto/istockphoto_test.go
```

reports all four added files.

Also, `memory.parseIstockID` can panic before downloading an image. `overload()` calls `d.memory.GetMemory(imageURL)` for every image `src`, and `parseIstockID` assumes every `https://` URL has an `m` query value:

```go
return urlParse.Query()["m"][0]
```

A local package test with this URL failed with `index out of range [0] with length 0`:

```text
https://media.istockphoto.com/id/123/photo/example.jpg?s=612x612&w=0&k=20&c=abc
```

So I think the next useful update would be formatting plus a defensive/fixture-backed test around image URLs that do not carry `m`.

PR #748, opened 2023-01-12
fix: enable select current element for struct slice unmarshal

xzvwang:
Once I met a html like this:
```
<div>
    <a href="url1">Name 1</a>
    <a href="url2">Name 2</a>
</div>
```
and I want extract the data as an array of name and url pair, so i write a selector:
```
[]*struct {
    Name string `selector:""`
    URL  string `selector:"" attr:"href"`
} `selector:"a"`
```
but the `Unmarshal` can not do it well, so i expand this function, add `unmarshalString` to handle if selector is empty string.


StantonMatt (commented):
I checked this against the current `origin/master` state. The behavior gap is still there: adding a regression that unmarshals `<a href="url1">Name 1</a>` into both `[]struct` and `[]*struct` fields with `selector:""` fails on master with empty `Name`/`URL` values.

The submitted branch's focused unmarshal tests pass, and `go test ./... -count=1` passes. I also applied the same idea onto current master in a scratch tree, using the current selection when the selector tag is present but empty; the focused regression and full suite passed there as well.

The branch still needs a refresh before merge: it conflicts in `unmarshal.go` because master already changed the string-selector path, and `gofmt -l unmarshal.go unmarshal_test.go` reports both touched files. So the fix still looks valid to me, but it needs a rebase plus formatting and a current-master regression test.

PR #749, opened 2023-01-16
Update CI configuration

WGH-:
Upstream supports 1.19 and 1.18 right now, but let's be generous and also include 1.17.

This should fix CI, since Colly can't be built properly on 1.15 and older, and it has been this way for a while.

Also enable CI runs for pull requests, as it's good to know whether they're ok or not, and GitHub Actions are free anyway.


asciimoo (approved):
Totally agree, useful changes, thanks!

PR #755, opened 2023-03-01
fixed fixCharset ignore not enough bug

lorentz-wu:
fixCharset ignore all non text content type

### DIFF
--- a/response.go
+++ b/response.go
@@ -75,12 +75,7 @@ func (r *Response) fixCharset(detectCharset bool, defaultEncoding string) error 
 	}
 	contentType := strings.ToLower(r.Headers.Get("Content-Type"))
 
-	if strings.Contains(contentType, "image/") ||
-		strings.Contains(contentType, "video/") ||
-		strings.Contains(contentType, "audio/") ||
-		strings.Contains(contentType, "font/") {
-		// These MIME types should not have textual data.
-
+	if !strings.Contains(contentType, "text/") 
 		return nil
 	}
 


response.go
```diff
@@ -75,12 +75,7 @@ func (r *Response) fixCharset(detectCharset bool, defaultEncoding string) error
 	}
 	contentType := strings.ToLower(r.Headers.Get("Content-Type"))
 
-	if strings.Contains(contentType, "image/") ||
-		strings.Contains(contentType, "video/") ||
-		strings.Contains(contentType, "audio/") ||
-		strings.Contains(contentType, "font/") {
-		// These MIME types should not have textual data.
-
+	if !strings.Contains(contentType, "text/") {
```
WGH-, line 78: I agree that the original check is too simplistic, but they're a lot of textual MIME types without `text/` that might have encoding problems as well. There are  `application/xhtml+xml`, `application/xml`, `application/atom+xml`, to name a few.

colly.go
```diff
@@ -462,6 +464,12 @@ func CheckHead() CollectorOption {
 	}
 }
 
+func FixCharset(f bool) CollectorOption {
+	return func(c *Collector) {
+		c.fixCharset = f
+	}
+}
+
```
WGH-, line 472: How is this different from `DetectCharset` option?

WGH- (changes requested):

colly.go
lorentz-wu: if fixCharset  set false,  user can fix charset onResponse

PR #763, opened 2023-04-10
Support websites redirecting to the same page when AllowURLRevisit is disabled

WGH-:
Some websites set a session cookie, and return a redirect to the same page instead of returning a response.

To illustrate this problem, this is how HTTP session might look like:

    GET / HTTP/1.1
    Host: 127.0.0.1:34931
    User-Agent: colly - https://github.com/gocolly/colly/v2
    Accept: */*
    Accept-Encoding: gzip

    HTTP/1.1 302 Found
    Content-Type: text/html; charset=utf-8
    Location: /
    Set-Cookie: session_id=1
    Date: Mon, 10 Apr 2023 23:29:29 GMT
    Content-Length: 24

    <a href="/">Found</a>.

    GET / HTTP/1.1
    Host: 127.0.0.1:34931
    User-Agent: colly - https://github.com/gocolly/colly/v2
    Accept: */*
    Cookie: session_id=1
    Referer: http://127.0.0.1:34931/
    Accept-Encoding: gzip

    HTTP/1.1 200 OK
    Date: Mon, 10 Apr 2023 23:29:29 GMT
    Content-Length: 12
    Content-Type: text/plain; charset=utf-8

    hello world

This fixes regression introduced in 0be3b71547adc5 by specifically bypassing revisit check if current redirect destination equals to the original one.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1334,7 +1334,12 @@ func (c *Collector) checkRedirectFunc() func(req *http.Request, via []*http.Requ
 			return fmt.Errorf("Not following redirect to %q: %w", req.URL, err)
 		}
 
-		if !c.AllowURLRevisit {
+		// allow redirects to the original destination
+		// to support websites redirecting to the same page while setting
+		// session cookies
+		samePageRedirect := normalizeURL(req.URL.String()) == normalizeURL(via[0].URL.String())
+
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
LGTM, thanks <3

Issue #762, opened 2023-04-07
User Agent update and Edge Browser UA addition


samirsss:
We have the Chrome, Firefox and  Opera UA which are older versions from 2021 which need to be updated and also the newer Edge can be used for a user agent.

Could these be updated/added to the latest version of colly?

samirsss:
bump please

guessi:
@samirsss I've update #765 for addressing this issue as you pin me at #605, but I'm not maintainer of the package. So you will need to wait for maintainer's review so that it could get merged.

samirsss:
@guessi - I understand it and really appreciate you doing this PR. Hopefully @asciimoo can review and approve the PR. I gave it a run and it looks good.

PR #765, opened 2023-04-22
Update User-Agent strings for 2023

guessi:
- Removed Firefox version earlier than Jun 1, 2022
- Removed Google Chrome version earlier than Jun 1, 2022
- Removed Microsoft Edge version earlier than Jun 1, 2022
- Removed Opera version earlier than Jan 1, 2023
- Updated Mac OS version strings (10.2.x ~ 13.3.x)
- Extended support for Android 12/13
- Updated build code for Nexus 10
- Remove legacy devices support
- Add support for Pixel 7
- Add support for Pixel 6
- Add support for Pixel 5
- Add support for Pixel 4

Fixes: #762

### DIFF
--- a/extensions/random_user_agent.go
+++ b/extensions/random_user_agent.go
@@ -52,6 +52,7 @@ var ffVersions = []float32{
 	110.0,
 	111.0,
 	112.0,
+	113.0,
 }
 
 var chromeVersions = []string{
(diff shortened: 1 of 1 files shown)


samirsss:
bump

guessi:
Just made an update again to support more mobile devices.

samirsss:
Bump @asciimoo

WGH-:
I personally think the code is too complicated though, just a static list of strings would be sufficient for the job. But let's be conservative for the time being.

guessi:
@WGH- I agree. I simply try to keep the old way  before we find a way to keep it update automatically.

Maybe it could be a static list generated by script and trigger monthly/quarterly with CI/CD job, that should be enough?

WGH-:
I think the main problem is that it may generate some weird unlikely OS version/browser version combinations, which may tip off bot detection algorithms.

For most purposes, just a few common UA strings should be more than enough.

guessi:
True 😂

That's the reason behind the scene why I introduce new UA strings for mobile.

PR #766, opened 2023-04-29
Refactor existing caching mechanism into pluggable system

### LINKED ISSUE #103: Pluggable cache system?
Hi,

Thanks for Colly! :)

I have a task with hundreds of thousands of pages, so obviously I am using Colly's caching, but it's basically 'too much' for the filesystem. (Wasted disk space, a pain to manage, slow to backup, etc)

I'd like to propose a pluggable cache system, similar to how you've made other Colly components.

Perhaps with an API like this:-

```go
type Cache interface {
	Init() error
	Get(url string) (*colly.Response, error)
	Put(url string, r colly.Response) error
	Remove(url string) error
}
```

...or...

```go
type Cache interface {
	Init() error
	Get(url string) ([]byte, error)
	Put(url string, data []byte) error
	Remove(url string) error
}
```

The first one won't be possible if you then wish to implement FileSystemCache in a subpackage to Colly though.

The reason I also need a Remove method is because one project has a site that sometimes serves maintenance pages, and whilst I can detect these, Colly currently has no method of saying stop, or of rejecting a page after processing. Obviously, the last thing I want part way through a crawl is to have my cache poisoned. But that's probably a separate issue, that I can live with if I can do the removal of bad pages myself.

If pluggable caches were to be implemented, I have existing code from another project that has a cache built using SQLite as a key-value store, compressing/decompressing the data with Zstandard (it's both surprisingly quick and super efficient on disk space), that I would happily port over. This can either become part of Colly, or a separate thing on my own Github.

I did start implementing this myself, but ran into a problem with how I went about it. (I followed the pattern you have established of having the separate components as subpackages, I then got bitten because my FileSystemCache couldn't easily reach into the Collector to get the CacheDir, I was trying to preserve existing behaviour / API compatibility. Maybe that's not an issue. Maybe these bits shouldn't be subpackages. Obviously once I started thinking things like that I figured it was best to discuss before/instead of proceeding any further.)

— Your thoughts?

unjello:
Create an interface `Cache`, similiar to `storage.Storage` that will allow for pluggable caching extensions. Caching subsystem is created with full access to Request and Response objects, to potentialy make more complex caching decisions based on both request and response headers. Because of the need to access those objects, `Cache` is not created in its own package like `storage`, but in root `colly`.

Closes #103


unjello:
@asciimoo would you happen to have a second to review this?

cache.go
```diff
@@ -0,0 +1,142 @@
...
+// Cache is an interface which handles caching Collector's responses
+// The default Cache of the Collector is the NullCache.
+// FileSystemCache keeps compatibility with non-pluggable caching in legacy
+// Collector. For this reason, one can set cache folder via `CACHE_DIR`
+// environment variable, or by passing CacheDir to NewCollector.
+// Collector's caching backend can be changed too by calling new method
+// Collector.SetCache
+type Cache interface {
+	// Init initializes the caching backend
+	Init() error
```
WGH-, line 36: Do we really need `Init()` as a part of the interface? This is not typical in Go. Can't we just assume that whateve `Cache` implementation is passed to colly, it's already initialized?

cache.go
```diff
@@ -0,0 +1,142 @@
...
+	// Get retrieves a previously cached response for the given request
+	Get(request *http.Request) (*Response, error)
+	// Put stores a response given request in cache
+	Put(request *http.Request, response *Response) error
+	// Close finalizes caching backend
+	Close() error
```
WGH-, line 42: What's the expected lifecycle of the `Cache` implementation? Is it tightly coupled to the `Collector`'s one, or it can be shared between several collectors, and something else must close it?

cache.go
```diff
@@ -0,0 +1,142 @@
...
+	dir := path.Join(c.BaseDir, hash[:2])
+	return dir, path.Join(dir, hash)
+}
+
+// Get returns an error for HTTP verbs other than GET and if request headers
+// specify `Cache-Control: no-cache`.
+func (c *FileSystemCache) Get(request *http.Request) (*Response, error) {
+	if request.Method != "GET" || request.Header.Get("Cache-Control") == "no-cache" {
+		return nil, ErrRequestNoCache
+	}
```
WGH-, line 104: Does this check has to be inside `Cache` implementation rather than somewhere in its caller? I think cacheability of a HTTP request doesn't really depend on where you store its response.

WGH-:
I certainly approve the efforts do decouple caching from the guts of colly.

cache.go
unjello: The main reason I've decided to go with `Init` is consistency - it may not be idiomatic for Go, but it is used by `colly` all over the place... `Collector`, `httpBackend`, etc. My other reason was the way I was thinking about backwards compatibility - i.e. passing `SetCollector` via `NewCollector`, but after some thinking, I may call `NewFileSystemCache` just as well. If you'd like me to remove `Init` just let me know.

cache.go
unjello: That's a very good question. The backwards-compatible `FileSystemCache` is tightly coupled with `Collector`, but the way to use other caches would be via `Collector.SetCache` only. Unfortunately in current form, where it's `.SetCache` who's calling `cache.Init` it kind of defeats the purpose. I will remove `.Init` from the `.SetCache` and document that this method should pass an initialised object, and a caller is responsible for closing. For the backwards-compatibility scenario, it should not matter too much. `Collector` does not have a `.Close` method, and `FileSystemCache` did not do any *closing* either.

cache.go
WGH-: > I may call NewFileSystemCache just as well. If you'd like me to remove Init just let me know.

Yeah, that sounds much better. Please do.

cache.go
WGH-: Okay, at least let's move this check into unexported function for the time being.

cache.go
WGH-: I thought about this a bit.

I believe if cache is set through the environment variable, collector's `Close` should call the cache's `Close` as well. If cache is initialized through `SetCache`, then it should not be closed. Moreover, if there's a cache previously initialized through the environment variable when `SetCache` is called, it should be closed before replacing it with the new one.

Do you agree?

PR #779, opened 2023-09-12
support for Request option model 

lingdor:
most cases , you maybe want add more information to url, for example the database id of url and others.at done of the task to associate to other record.


lingdor (commented):
the code is simple, but can give collyer more easily. please code minister review the code ,3ks

PR #782, opened 2023-09-18
support for Request option model 

lingdor:
most cases , you maybe want add more information to url, for example the database id of url and others.at done of the task to associate to other record.


WGH-:
Couldn't you already achieve the same with `Queue.AddRequest`?

StantonMatt (commented):
I checked this because of the earlier question about whether `Queue.AddRequest` already covers the same use case.

The PR head `f34f923` builds locally:

```text
go test . -run TestNewRequest -count=1 -v
go test ./queue -count=1
go test ./... -count=1
```

I also verified the existing API path on current Colly: constructing a `*colly.Request` with `Ctx` and `Headers`, then passing it to `Queue.AddRequest`, carries both values through `Queue.Run` in a local scratch test. So the main behavior in the updated example can already be expressed without adding an `AddURL` options API.

One issue with the PR as written is the new queue example: it stores the loop index in the context (`ctx.Put` with key `num` and integer value `i`), then queues the URL with `WithRequestContext(ctx)`.

Because queued requests are marshaled through JSON, that integer round-trips as `float64`. If a callback later calls `Context.Get` for the `num` key, it panics at `context.go:60` because `Get` type-asserts to `string`.

I reproduced that with a local scratch test using the PR API:

```text
panic: interface conversion: interface {} is float64, not string
```

If this API is still wanted as a convenience wrapper around `AddRequest`, I think it needs tests for the `AddURL(...opts)` path and the example should either store a string context value or use `GetAny` for non-string values.

PR #785, opened 2023-10-03
Replaced deprecated io/ioutil package with proper implementations of …

zachary-walters:
…io/os packages.

As of Go v1.16, the io/ioutil package has been deprecated and replaced with io and os implementations. https://go.dev/doc/go1.16#ioutil


WGH-:
I don't think these deprecated aliases will go away any time before Go 2 (so never), but it's a good idea to clean up the code from time to time nevertheless. Thanks!

WGH- (approved):

Issue #777, opened 2023-08-23
Bug encoding cyrillic windows-1251


Dinver:
Hi! When I try to recognize the encoding on sites with windows-1251, I get:
2023/08/23 21:45:10 ÑÄÎ «Ïðîìåòåé» | ÎÎÎ «Âèðòóàëüíûå òåõíîëîãèè â îáðàçîâàíèè»
2023/08/23 21:45:10 Ýëåêòðîííûå êóðñû
2023/08/23 21:45:10 Ïðîäóêòû

Example:
`
package main

import (
	"log"

	"github.com/gocolly/colly"
)

func main() {
	c := colly.NewCollector(
		colly.DetectCharset(),
		colly.Async(true),
	)
	c.OnHTML("title", func(e *colly.HTMLElement) {
		title := e.Text
		log.Println(title)
	})
	c.OnHTML("a[href]", func(e *colly.HTMLElement) {
		title := e.Text
		log.Println(title)
	})

	c.OnHTML("img, script, link", func(e *colly.HTMLElement) {
		title := e.Attr("alt")
		log.Println(title)
	})

	c.Visit("https://prometeus.ru/")
	c.Wait()
}

`

colly.DetectCharset() / c.DetectCharset = true - does not working.

blagoySimandov:
Pretty sure this is not a problem with Colly but with the terminal. Most terminals do not support cyrillic output. If you put everything in a database everything should look fine (I've crawled cyrillic pages before and I know that it works). But in case you really need to have the output displayed in the terminal try using something like Windows PowerShell ISE - it has a fairly good support for displaying Unicode.

Dinver:
> Pretty sure this is not a problem with Colly but with the terminal. Most terminals do not support cyrillic output. If you put everything in a database everything should look fine (I've crawled cyrillic pages before and I know that it works). But in case you really need to have the output displayed in the terminal try using something like Windows PowerShell ISE - it has a fairly good support for displaying Unicode.

It's not about the terminal, this example is just to reproduce the error. The on API data is also sent incorrectly.

WGH-:
Yeah, I can reproduce it with colly/v2, too

Dinver:
Solved the problem, by adding a check meta[http-equiv='Content-Type'] in body, in the absence of a "charset" but with "text/html" in the header. I don't know if this is the correct approach, but it solves the problem.

response.go:
```
package colly

import (
	"bytes"
	"fmt"
	"io/ioutil"
	"mime"
[...]

PR #788, opened 2023-10-09
Encoding detection fix

blagoySimandov:
Connected to issue #777 "HTML encoding is not autodetected properly". I removed the current gocolly encoding detection, which through tests showed to be unreliable when detecting Cyrillic encodings, and in place of it used the built-in function _DetermineEncoding_ from the _charset_ package.


response.go
```diff
@@ -88,12 +86,8 @@ func (r *Response) fixCharset(detectCharset bool, defaultEncoding string) error
...
 			return nil
 		}
-		d := chardet.NewTextDetector()
-		r, err := d.DetectBest(r.Body)
-		if err != nil {
-			return err
-		}
-		contentType = "text/plain; charset=" + r.Charset
+		_, nameOfEncoding, _ := charset.DetermineEncoding(r.Body, contentType) //name of charset/encoding
+		contentType = "text/plain; charset=" + nameOfEncoding
```
WGH-, line 90: `charset.DetermineEncoding` returns the detected encoding, so I think this entire function can be greatly simplified and replaced with rougly this:

```go
enc, _, certain := charset.DetermineEncoding(r.Body, contentType)
if !certain && !detectCharset {
	return nil
}
var err error
r.Body, err = ioutil.ReadAll(enc.NewDecoder().Reader(bytes.NewReader(r.Body)))
return err
```

response.go
WGH-, line 1: This needs a unit test.

WGH- (changes requested):
Thanks! This needs a couple of changes, though (see above).

And please clean up your commit history (or I will do it myself before the merge)

StantonMatt (commented):
I checked this on a current `master` merge. The branch still applies cleanly, and the existing suite passes there:

```text
go test ./... -count=1
git diff --check origin/master
```

`gofmt -l response.go` still reports `response.go`, so the submitted diff needs formatting.

I also added a local-only regression for the behavior this PR is trying to fix. With `DetectCharset`, current `master` misdecodes HTML bodies that declare their encoding through a meta tag but have only `Content-Type: text/html` in the response header:

```text
shift-jis meta:    expected the UTF-8 Japanese body text, but got mojibake bytes rendered as `\x22\xfa-\x7b\x8c\xea`
windows-1251 meta: expected the UTF-8 Cyrillic body text, but got mojibake bytes rendered as `\xcf\xf0\xe8\xe2\xe5\xf2`
```

The same test passes on this PR merged into current `master`, because `charset.DetermineEncoding` sees the HTML meta charset before decoding.

So the behavior change still looks useful to me. I think the remaining merge blockers are the ones already called out in review: format `response.go`, simplify around the `charset.DetermineEncoding` result, and commit a unit test for this no-header-charset/meta-charset case.

PR #746, opened 2023-01-05
Don't decompress gzip if data doesn't look like gzip

WGH-:
See #745 for more information.

Closes #745

PR #792, opened 2023-11-03
Fix attempt to read gzip content from responses that cannot contain content

seedifferently:
Colly currently returns an `EOF` error on responses that contain a `Content-Encoding: gzip` header, but no body content.

This happens due to the initialization of the gzip reader [here](https://github.com/gocolly/colly/blob/3c987f1982edbb5ba8876eef56dd35e1ff05932a/http_backend.go#L205C1-L205C1) attempting to parse gzip header data from the response body, even though the body is empty.

RFC 9110 section 15 states that [1xx](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.2), [204](https://www.rfc-editor.org/rfc/rfc9110.html#name-204-no-content), and [304](https://www.rfc-editor.org/rfc/rfc9110.html#name-304-not-modified) responses cannot contain content. However, some servers will still send a `Content-Encoding` header.

This is my best guess at how to instruct colly not to reach for the gzip reader in these scenarios (the non-gzip reader gracefully handles empty bodies).

### DIFF
--- a/colly_test.go
+++ b/colly_test.go
@@ -131,6 +131,11 @@ func newUnstartedTestServer() *httptest.Server {
 		w.Write([]byte("ok"))
 	})
 
+	mux.HandleFunc("/204_enc_gzip", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Encoding", "gzip")
+		w.WriteHeader(204)
+	})
+
 	mux.HandleFunc("/500", func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "text/html")
 		w.WriteHeader(500)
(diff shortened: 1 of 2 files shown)


WGH-:
I think my #746 solves it in more general case. The problem is the PR got stalled on review, as I can't self-approve and merge

seedifferently:
@WGH- thanks, I had not seen that PR. However, if you copy my test case over there you will see it fail because the `Peek()` still returns an `EOF`, so some additional checking would be needed to determine if the `EOF` is OK to ignore and move on from.

WGH-:
That's a good catch. I'll update my PR, and credit you.

seedifferently:
Thanks @WGH-.
Closing this in favor of  #746.

PR #793, opened 2023-11-12
Update User-Agent strings for Firefox/Opera/Chrome/Edge

mtskg:
* Following the same pattern as in 75c67a68d08d9371dfcbf7083213cefcd9082c8c which also included developer and beta updates of Opera.

### DIFF
--- a/extensions/random_user_agent.go
+++ b/extensions/random_user_agent.go
@@ -56,6 +56,12 @@ var ffVersions = []float32{
 	111.0,
 	112.0,
 	113.0,
+	114.0,
+	115.0,
+	116.0,
+	117.0,
+	118.0,
+	119.0,
 }
 
 var chromeVersions = []string{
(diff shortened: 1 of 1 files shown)


houseme (approved):

houseme:
@asciimoo

asciimoo:
Could you please update the PR to the latest versions and keep only the version strings that are supported by their devs?

houseme:
> Could you please update the PR to the latest versions and keep only the version strings that are supported by their devs?

@mtskg

StantonMatt (changes requested):
I checked the current head `25dfb510` against current `master`.

Local checks:

```text
go test ./extensions -count=1
go test ./... -count=1
```

Those pass. `gofmt -l extensions/random_user_agent.go` still reports `extensions/random_user_agent.go`.

I think the update still needs another pass before it matches the maintainer request to use latest supported version strings:

- The added lists stop at Firefox/Chrome 119. Current official feeds I checked report Firefox `151.0.3` / DevEdition `152.0b6` from `https://product-details.mozilla.org/1.0/firefox_versions.json`, and Chrome stable Windows `149.0.7827.54` from ChromiumDash.
- One added Chrome entry is release-note shorthand rather than a UA-safe version: `114.0.5735.90/91`. Since `genChromeUA` emits it as `Chrome/%s`, that can produce `Chrome/114.0.5735.90/91 Safari/...`. A local-only guard test over `chromeVersions` failed on that exact value.

So I would refresh the data against current supported versions, split/remove shorthand entries, and add a small guard test for UA-safe Chrome version strings.

PR #800, opened 2024-01-04
Allow access to unsanitized file name

j0hax:
Hello there! I've come across a situation where I have to save a file with a "double" extension (`*.kepub.epub`), and the current implementation of (r *[Response](https://pkg.go.dev/github.com/gocolly/colly/v2#Response)) FileName() purposefully breaks that by always calling `SanitizeFileName()`.

I propose to provide direct access to an unsanitized file name with the method `RawFileName()`. This shouldn't affect compatibility and allows for accessing information similar to [URL package](https://pkg.go.dev/net/url#URL)'s `RawPath`, `RawQuery`, etc.

### DIFF
--- a/response.go
+++ b/response.go
@@ -49,17 +49,23 @@ func (r *Response) Save(fileName string) error {
 	return os.WriteFile(fileName, r.Body, 0644)
 }
 
+// RawFileName returns the unsanitized file name parsed from "Content-Disposition"
+// header or from URL. Be aware that the name may contain unsafe characters.
+func (r *Response) RawFileName() string {
+	_, params, err := mime.ParseMediaType(r.Headers.Get("Content-Disposition"))
+	if fName, ok := params["filename"]; ok && err == nil {
+		return fName
+	}
+	if r.Request.URL.RawQuery != "" {
+		return fmt.Sprintf("%s_%s", r.Request.URL.Path, r.Request.URL.RawQuery)
+	}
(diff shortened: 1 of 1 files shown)


WGH-:
Ehm, perhaps, sanitizing double extension away was a mistake in the first place.

StantonMatt (commented):
I checked this locally. The change is backwards-compatible for `FileName()` because it now just sanitizes `RawFileName()`, and the full Go suite passes on the PR branch.

What I ran:

```sh
GOCACHE=.codex-tmp/go-cache GOTMPDIR=.codex-tmp/go-tmp GOMODCACHE=.codex-tmp/go-mod go test ./... -count=1
git diff --check upstream/master...HEAD
```

Both passed. GitHub reports no checks for this branch.

I also used a local-only test with `books/book.kepub.epub` and a `Content-Disposition` filename like `../../unsafe.kepub.epub`. That confirmed the tradeoff: `RawFileName()` preserves the exact header/path-derived name, including path separators and unsafe segments, while `FileName()` still returns a safe filename and still folds the double extension into names like `books_book_kepub.epub`.

So this does solve the caller need without changing existing `FileName()` behavior. The maintainer question above still seems important, though: if the double-extension folding is itself the bug, a smaller change may be to adjust `SanitizeFileName()`/`FileName()` so safe double extensions survive. If both APIs are desired, I’d add regression tests for `RawFileName()` and the existing sanitized behavior so the safety distinction is explicit.

PR #808, opened 2024-03-25
Implement content sniffing for HTML parsing

WGH-:
Web pages can be served without `Content-Type` set, in which case browsers employ content sniffing. Do the same here, in Colly.

While we're at it, change the Content-Type check to something stricter than mere "html" substring match.


WGH-:
Welp, `strings.Cut` appeared only in Go 1.18. Instead of rewriting it the old way I decided to drop old Go versions (#810).

PR #810, opened 2024-03-25
Test on Go 1.22, drop old versions

WGH-:
The Go team officially supports the two latest releases, so we're even being generous here.

Go 1.17 notably lacks useful strings.Cut function.


asciimoo (approved):
Thanks!

PR #811, opened 2024-03-29
Fix more cases of pages redirecting to themselves

WGH-:
This was "fixed" in b4ca6a7a17 (#763), but the fix turned out to be incomplete.

That fix only allowed redirects leading to the same URL as the original destination, and didn't take into account more complicated cases. Such as, for example:

 * www.example.com
 * example.com
 * (set cookie)
 * example.com

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1463,12 +1463,26 @@ func (c *Collector) checkRedirectFunc() func(req *http.Request, via []*http.Requ
 			return fmt.Errorf("Not following redirect to %q: %w", req.URL, err)
 		}
 
-		// allow redirects to the original destination
-		// to support websites redirecting to the same page while setting
-		// session cookies
-		samePageRedirect := normalizeURL(req.URL.String()) == normalizeURL(via[0].URL.String())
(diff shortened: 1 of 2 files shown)


melroy89:
@asciimoo Can you maybe add WGH to your repository with merge permissions? I believe you are unable to maintain the project. Maybe add some people to help you out.

WGH-:
I can merge, I just can't review my own MR :)

melroy89 (approved):
Hope this helps. Approved. However, I'm not part of the team which has write access. This is not good. We are blocked!

WGH-:
Unfortunately, no: I need a review by someone with write access

melroy89:
> Unfortunately, no: I need a review by someone with write access

In that case we need more maintainers. Since this is slowing down development by a lot!

I'm also running big open source projects. And I embraced the C4 specification.

asciimoo:
I've temporarily removed the required approval to merge pull request. @WGH- you can merge contributions without my approval.

PR #818, opened 2024-06-14
Add wikipedia example

sarahegler:
[Wikipedia Six Degrees of Separation](https://www.sixdegreesofwikipedia.com/?source=Coldplay&target=Jack%20Sparrow) Example


StantonMatt (commented):
I checked this locally because the branch is dirty but the example itself is small.

On the PR head 362752f, these pass:

- go test ./_examples/wikipedia
- go test ./... -count=1

The current conflicts are only from go.mod and go.sum. Current master already has newer dependency versions than this branch in the touched module entries, so I also tried applying only _examples/wikipedia/main.go onto current master; go test ./_examples/wikipedia still passes there without any module changes.

The remaining mechanical issue is formatting: gofmt -l _examples/wikipedia/main.go reports the new file. I think this would be cleaner as just the formatted example file, rebased onto current master, with the go.mod/go.sum changes dropped.

PR #819, opened 2024-06-15
Don't pass unused parameters in fetch method

roarc0:


StantonMatt (commented):
I checked this locally. The cleanup looks behavior-neutral: the removed `fetch` arguments are already available through the constructed `*http.Request`, and `Request` is still populated from `req.URL`, `req.Header`, `req.Host`, the method, body, context, and depth.

Validation:

```sh
go test ./... -count=1
```

Both passed. GitHub reports no hosted checks for this PR branch.

PR #832, opened 2024-11-15
doc: mention that OnScraped executes after OnHTML and OnXML

aladmit:
In my project, I have `OnHTML` and `OnXML` callbacks, and I got confused when I read in the documentation that `OnScraped` would be executed after `OnHTML`. To figure out what would happen with `OnXML`, I had to dive into the source code.

This PR makes the `OnScraped` description a little clearer.


asciimoo (approved):
Thanks!

PR #838, opened 2025-02-25
Add headers/ua to robots.txt request

### LINKED ISSUE #837: robots.txt checking does not use provided user agent
https://github.com/gocolly/colly/blob/99b7fb1b87d1491578f1fa9c222836341f533d75/colly.go#L832

The line above is how robots.txt files are requested. Because a simple `.Get` is used, this request isn't made with the user agent header the user may have set.

In my case this is a critical issue because one of the hosts I am making requests to blocks the default Go user agent, `Go-http-client/2.0`, for whatever reason. So this request fails, and then the initial crawl request fails too.

makew0rld:
Fixes #837

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -829,7 +829,31 @@ func (c *Collector) checkRobots(u *url.URL) error {
 
 	if !ok {
 		// no robots file cached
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Thank you!

PR #841, opened 2025-03-27
fix #835: calling a callback detach while executing a callback might cause callbacks to be skipped or executed twice

### LINKED ISSUE #835: Calling OnHTMLDetach while inside a OnHTML will cause the next callback to be skipped
If I detach a HTML Callback while I am inside it, it will be removed from the slice of htmlCallbacks, unfortunately this will cause the next callback to be skipped while colly is ranging over the callbacks.

This is the function to detach ([link](https://github.com/gocolly/colly/blob/99b7fb1b87d1491578f1fa9c222836341f533d75/colly.go#L950))
``` 
func (c *Collector) OnHTMLDetach(goquerySelector string) {
        ...
		c.htmlCallbacks = append(c.htmlCallbacks[:deleteIdx], c.htmlCallbacks[deleteIdx+1:]...)
	...
}
```

And if it is called while colly is still ranging over the callbacks ([link](https://github.com/gocolly/colly/blob/99b7fb1b87d1491578f1fa9c222836341f533d75/colly.go#L1155))
``` 
for _, cc := range c.htmlCallbacks {
		i := 0
		doc.Find(cc.Selector).Each(func(_ int, s *goquery.Selection) {
			...
				cc.Function(e)
			}
		})
	}
```

the next callback will be skipped because the slice shrinks by one while it is iterated over.

horstf:
I've added a fix for the issue #835 of calling `html` and `xml` detaches while inside a `html` or `xml` callback. Previously this caused the slice to be modified while iterating over it, which might cause other callbacks to be skipped or to be executed twice (see [this playground](https://go.dev/play/p/5MDOu-Eo5LQ) for a short demo of a bug when modifying a slice while ranging over it). 

This PR changes the callback execution logic to introduce a parameter `active` and track which callbacks have `active == true` and only execute those. Calling a detach will set `active` to `false` for the callback, at a later step we remove all inactive callbacks.

This PR also adds a test for this issue.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,7 @@
+{
+    // Verwendet IntelliSense zum Ermitteln möglicher Attribute.
+    // Zeigen Sie auf vorhandene Attribute, um die zugehörigen Beschreibungen anzuzeigen.
+    // Weitere Informationen finden Sie unter https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": []
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,0 @@
-{
-    // Verwendet IntelliSense zum Ermitteln möglicher Attribute.
-    // Zeigen Sie auf vorhandene Attribute, um die zugehörigen Beschreibungen anzuzeigen.
-    // Weitere Informationen finden Sie unter https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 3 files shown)


asciimoo:
Great, thanks!

asciimoo:
@horstf Seems like the PR introduced data race warnings in the tests. Could you please check it? You can reproduce the test errors by running `go test -race ./...` in the project root

horstf:
I will check it out

PR #844, opened 2025-03-29
Add chromedp backend

markwinter:
I rebased this but haven't checked it too thoroughly.

This adds a HTTPDriver interface and implements chromedp as one of the implementations.


t-gaudam:
Is there any update on this? are we planning to merge this

asciimoo:
I'd like to merge it if we receive 1-2 positive feedbacks that it is stable and works as excepcted.

markwinter:
I think in the latest rebase it broke, it hangs someplace.

I don't currently have capacity to work on it at the moment. Anyone else is free to work on top of this until I can get to it

StantonMatt (commented):
I took a pass on this because the maintainer thread asked for stability feedback and the author noted that the latest rebase hangs/breaks.

On the current PR head `0e723af`, `go test ./... -count=1 -timeout=3m` does not reach the tests yet:

```text
go: updates to go.mod needed; to update it:
	go mod tidy
```

After a local `go mod tidy`, the next blocker is a build error in the robots path:

```text
./colly.go:848:56: too many arguments in call to robotstxt.FromResponse
	have (int, []byte)
	want (*http.Response)
```

That part looks like a small API mismatch from switching robots fetching through the new `HTTPDriver` response shape. `robotstxt.FromStatusAndBytes(resp.StatusCode, resp.Body)` matches the values the branch now has.

I also reproduced why the chromedp backend was not usable in a local smoke test. With a temporary `httptest` page that writes `<h1 id="dynamic">rendered</h1>` from JavaScript, `CDPDriver.Do()` either returned `Could not find node with given id (-32000)` or returned rendered HTML that Colly did not parse through `OnHTML`. Two small local changes fixed that smoke path for me:

```go
chromedp.Navigate(request.URL.String()),
chromedp.WaitReady("html"),
chromedp.OuterHTML("html", &body),
```

and returning `Content-Type: text/html` instead of `html` from the Colly `Response`.

With those local changes plus the tidy update, these passed locally:

```text
go test . -run TestCodexChromeDriverSmoke -count=3 -timeout=60s -v
go test ./... -count=1 -timeout=3m
```

The smoke test did still log chromedp/cdproto messages about `network.IPAddressSpace` value `Loopback` with my installed Chrome, but the rendered DOM was returned and `OnHTML` saw the dynamic element. So a future cdproto/chromedp refresh may reduce protocol noise, but the basic backend path worked here with the PR dependency versions once the response extraction and content type were adjusted.

The only visible GitHub status on the PR is License Compliance, which is green; no code CI is reported on the branch.

StantonMatt:
I checked the current head `0e723af` locally. It fails before tests on the robots.txt path:

```text
go test . -run TestSetHTTPDriver -count=1 -v
```

```text
.\colly.go:848:56: too many arguments in call to robotstxt.FromResponse
        have (int, []byte)
        want (*http.Response)
```

That looks like a small mismatch from routing the robots fetch through `HTTPDriver.Do`, which returns Colly's `Response` shape. This local change got the branch compiling for me:

```go
robot, err = robotstxt.FromStatusAndBytes(resp.StatusCode, resp.Body)
```

I also had to run `go mod tidy`; with Go 1.24.1 that changes `go 1.23.0` to `go 1.24` and removes stale indirect/sum entries.

After those local edits, these passed:

```text
go test . -run TestSetHTTPDriver -count=1 -v
go test ./... -count=1
git diff --check
```

I did not reproduce a `go test ./...` hang after those fixes. I could not validate the actual chromedp runtime path on this machine because there is no Chrome/Chromium/Edge binary installed, so this only covers build and non-browser tests.

PR #846, opened 2025-04-26
docs: correct comment in SetProxyFunc to refer to *http.Transport

Marv963:
### What
The comment above Collector.SetProxyFunc said it only overrides the transport
if it “is not http.RoundTripper”, but the code actually looks for *http.Transport.
This PR updates the comment to correctly state that we override when the
existing Transport is not of type *http.Transport.

### Why
- Keeps documentation in sync with implementation.
- Avoids confusion for future contributors reading the code.

### How to test
This is purely a comment change; no behavior is affected.


asciimoo (approved):
Thanks

Issue #822, opened 2024-07-25
Multipart body should not be `map[string][]byte`


abiriadev:
First of all, thanks for the wonderful project! `colly` has saved our team a lot of time!!

## Context

ref: #8, #33

According to [RFC7578 section 4.3](https://www.rfc-editor.org/rfc/rfc7578#section-4.3):

> [4.3](https://www.rfc-editor.org/rfc/rfc7578#section-4.3).  Multiple Files for One Form Field
> 
> The form data for a form field might include multiple files.
> [RFC2388] suggested that multiple files for a single form field be transmitted using a nested "multipart/mixed" part.  This usage is deprecated.
> 
> <ins>To match widely deployed implementations, multiple files MUST be sent by supplying each file in a separate part but all with the same "name" parameter.</ins>
> 
> Receiving applications intended for wide applicability (e.g., multipart/form-data parsing libraries) SHOULD also support the older method of supplying multiple files.

and this practice is unsurprisingly common, and I am facing the exact same case.

## The issue

The `name` field does not have to be unique. There are few common cases when a duplicated `name` field is required (e.g., when uploading an array of files), and this case should be properly covered.

https://github.com/gocolly/colly/blob/99b7fb1b87d1491578f1fa9c222836341f533d75/colly.go#L551-L559

https://github.com/gocolly/colly/blob/99b7fb1b87d1491578f1fa9c222836341f533d75/colly.go#L1461-L1469

Unfortunately, the current implementaion accepts `map[string][]byte`, which enforces `name` to be unique.

## Suggestion

Maybe we can accept `[]Subpart` so that:

1. The order of subparts is guaranteed
2. `filename` and other metadata can be optionally included
3. Duplicate `name` fields are allowed

and so on.

I would love to hear your opinion! If you think this is feasible, I will start working on it.

jonesrussell:
Howdy, something like this https://github.com/gocolly/colly/pull/849 ?

abiriadev:
> Howdy, something like this [#849](https://github.com/gocolly/colly/pull/849) ?

Looks great!

asciimoo:
Looks great, indeed. What is the reason for it is being a draft? I'd be happy to merge it - @jonesrussell could you please remove the draft label and rebase it to the latest HEAD?

PR #849, opened 2025-05-08
Multipart body should not be map[string][]byte

jonesrussell:
Just fartin' around, saw https://github.com/gocolly/colly/issues/822 and wanted to see what Cursor thought.

### DIFF
--- a/_examples/multipart/multipart.go
+++ b/_examples/multipart/multipart.go
@@ -10,17 +10,17 @@ import (
 	"github.com/gocolly/colly/v2"
 )
 
-func generateFormData() map[string][]byte {
+func generateFormData() []colly.FormField {
 	f, _ := os.Open("gocolly.jpg")
 	defer f.Close()
 
 	imgData, _ := io.ReadAll(f)
 
-	return map[string][]byte{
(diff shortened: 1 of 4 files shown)


StantonMatt (commented):
I checked this against #822 because the linked issue has maintainer interest and the branch is still waiting on undraft/rebase.

On `16df369`, the implementation does solve the duplicate-name shape that `map[string][]byte` could not represent. In addition to the committed focused tests, I ran a local-only smoke test that posted two `FormField{Name: "upload", ...}` entries and verified `r.MultipartForm.Value["upload"] == []string{"first", "second"}` after `ParseMultipartForm()`.

Local checks on the PR branch:

```text
go test . -run 'TestCollectorPostMultipart' -count=1 -v
go test ./... -count=1
```

Those passed. The only visible GitHub status is License Compliance, which is green; no code CI is reported for the branch.

Before undrafting/rebasing, I would add a committed regression for the duplicate-name case from #822. The current new tests cover multipart parsing, files, and different field names, but not the main same-name/multiple-parts behavior that motivated the API change.

PR #851, opened 2025-05-22
ensure Referer header is carried around

carljia999:
Add HTTP header Referer in Request.Visit function so that we know the crawling order of HTTP requests. This is to mimic the default behaviour of standard browsers.


asciimoo:
We have an extension that does exactly this: https://github.com/gocolly/colly/blob/master/extensions/referer.go

carljia999:
Thank you very much for the information.

PR #852, opened 2025-05-29
Add cache expiration option

### LINKED ISSUE #834: Stale Cache system
Hey I'm using colly for a personal project of mine, and I'd like to not spam the web page I'm scraping.
I saw there was a `CacheDir` function to colly, but what happens if the webpage changes? Maybe it has some new data.

How can I set a timeout that a cached page is stale after a day or so, and colly would automatically fetch the current one?

nebojsaj1726:
Resolves #834.

This pull request adds support for a new CacheExpiration option to the Collector struct, allowing users to set a duration after which cached responses are considered stale and will be refreshed.

### DIFF
--- a/_examples/coursera_courses/coursera_courses.go
+++ b/_examples/coursera_courses/coursera_courses.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"strings"
+	"time"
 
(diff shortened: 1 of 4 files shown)


asciimoo (approved):
LGTM, thanks!

PR #858, opened 2025-06-17
add IsAbort for Request

Shinku-Chen:
Sometimes, it is necessary to check whether the request was aborted in previous callbacks within the callback function, and determine if the current request is aborted.


asciimoo (approved):
Thanks

PR #859, opened 2025-06-17
CacheExpiration should be copied when cloned

knyghte:
Makes sense to me that CacheExpiration should be copied when you copy a collector


asciimoo (approved):
Indeed, thanks!

PR #864, opened 2025-10-11
refactor: use the built-in max to simplify the code

stemlaud:
In Go 1.21, the standard library includes built-in [max](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.

### DIFF
--- a/http_backend.go
+++ b/http_backend.go
@@ -66,11 +66,7 @@ type LimitRule struct {
 
 // Init initializes the private members of LimitRule
 func (r *LimitRule) Init() error {
-	waitChanSize := 1
-	if r.Parallelism > 1 {
-		waitChanSize = r.Parallelism
-	}
-	r.waitChan = make(chan bool, waitChanSize)
+	r.waitChan = make(chan bool, max(r.Parallelism, 1))
 	hasPattern := false
 	if r.DomainRegexp != "" {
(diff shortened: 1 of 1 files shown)


http_backend.go
```diff
@@ -66,10 +66,7 @@ type LimitRule struct {
 
 // Init initializes the private members of LimitRule
 func (r *LimitRule) Init() error {
-	waitChanSize := 1
-	if r.Parallelism > 1 {
-		waitChanSize = r.Parallelism
-	}
+	waitChanSize := max(r.Parallelism, 1)
 	r.waitChan = make(chan bool, waitChanSize)
```
asciimoo: We don't need `waitChanSize` variable anymore in this case.

asciimoo (changes requested):
I'd argue that it **greatly** simplifies the codebase, but thanks, it's better this way. Please change according to my note, then it is ready to merge.

http_backend.go
stemlaud: Modified. Please review again.

stemlaud:
> I'd argue that it **greatly** simplifies the codebase, but thanks, it's better this way. Please change according to my note, then it is ready to merge.

Modified. Please review again.

asciimoo (approved):
Thanks!

PR #866, opened 2026-01-05
Refactor parseSettingsFromEnv to avoid magic numbers

tieubaovi0217:
This PR simply refactors the parseSettingsFromEnv method in the collector by introducing a constant string for the environment variable prefix and removing magic numbers from its implementation.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -214,6 +214,9 @@ const (
 	CheckRevisitKey
 )
 
+// The prefix for environment variables of Colly settings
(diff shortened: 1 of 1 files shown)


asciimoo (commented):
Why is `EnvVariablePrefix` public?

tieubaovi0217:
> Why is `EnvVariablePrefix` public?
Thank @asciimoo , there's no need to make it as public, I updated it as private.

asciimoo (approved):
Thanks!

PR #867, opened 2026-02-17
Fix #745: Don't attempt gzip decompression of non-2xx responses for .xml.gz URLs

### LINKED ISSUE #745: Attempting to fetch nonexisting /sitemap.xml.gz results in "gzip: invalid header"
```go
package main

import (
	"log"
	"net/http"

	"github.com/gocolly/colly/v2"
)

func main() {
	// we have to disable Accept-Encoding: gzip,
	// because the remote server might send compressed
	// 404 page, avoiding the bug
	tr := &http.Transport{
		DisableCompression: true,
	}

	c := colly.NewCollector()
	c.WithTransport(tr)

	c.OnError(func(resp *colly.Response, err error) {
		log.Printf("err=%v", err)
	})

	c.Visit("https://example.org/sitemap.xml.gz")
}
```

Results in:
```
2023/01/05 22:30:27 err=gzip: invalid header
```

rodionsteshenko:
## Problem

When visiting a URL ending with `.xml.gz` that returns a non-2xx response (e.g., a 404 HTML page), colly unconditionally tries to decompress the response body as gzip. This causes `gzip.NewReader` to fail with `gzip: invalid header`, masking the actual HTTP error.

## Fix

Added a status code check (200-299) to the URL-path-based gzip detection condition. Only successful responses are decompressed when the detection is based solely on the URL ending with `.xml.gz`. Content-Encoding and Content-Type based gzip detection remain unchanged, as those indicate the server explicitly encoded the response.

## Test

Added `TestIssue745GzipURLWith404Response` that:
- Sets up a test server returning a 404 HTML page at `/sitemap.xml.gz`
- Verifies the response is treated as a normal HTTP 404 error, not a gzip decompression failure

The test **fails** without the fix (gets `gzip: invalid header`) and **passes** with the fix.

Tested locally on macOS ARM (Apple Silicon). Full test suite passes.

### DIFF
--- a/colly_test.go
+++ b/colly_test.go
@@ -281,6 +281,15 @@ y">link</a>
 		}
 	})
 
+	mux.HandleFunc("/sitemap.xml.gz", func(w http.ResponseWriter, r *http.Request) {
+		// Return a 404 HTML page for a non-existent .xml.gz URL.
+		// This simulates the scenario in issue #745 where a server
+		// returns an HTML error page for a missing gzipped sitemap.
+		w.Header().Set("Content-Type", "text/html")
+		w.WriteHeader(404)
+		w.Write([]byte(`<!DOCTYPE html><html><body><h1>404 Not Found</h1></body></html>`))
+	})
+
 	return httptest.NewUnstartedServer(mux)
 }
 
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
Thanks

PR #869, opened 2026-04-13
fix: use precise XML MIME type matching to avoid parsing .xlsx files

### LINKED ISSUE #790: handleOnXML tries to parse`.xlsx` files
The handleOnXML function attempts to parse responses with the content-type `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`. This is because the function looks for any mention of [xml in the content type](https://github.com/gocolly/colly/blob/9401ae4acc5d2155e0ee09fa71eef4d09d2e412a/colly.go#L1186). This results in a parse error when `xmlquery.Parse()` is called (For example: `encoding/xml.SyntaxError {Msg: "illegal character code U+0003", Line: 1}).

[XLSX files packaged as a zip](https://www.loc.gov/preservation/digital/formats/fdd/fdd000398.shtml#:~:text=The%20XLSX%20format%20uses%20the,a%20rectangular%20grid%20of%20cells.) - so can't be directly parsed as XML. 

It would be ideal to not try and parse these files, possibly by being more explicit in which content-types we consider to be XML.

Yanhu007:
Fixes #790

## Problem

`handleOnXML` uses `strings.Contains(contentType, "xml")` to detect XML responses. This incorrectly matches OOXML MIME types like:

- `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` (.xlsx)
- `application/vnd.openxmlformats-officedocument.wordprocessingml.document` (.docx)

These are ZIP archives, not XML, causing `xmlquery.Parse()` to fail with errors like `encoding/xml.SyntaxError`.

## Fix

Replace the substring match with precise MIME type checks per [RFC 7303](https://datatracker.ietf.org/doc/html/rfc7303):

- `text/xml`
- `application/xml`
- `*+xml` suffix (e.g. `application/rss+xml`, `application/atom+xml`)

This correctly excludes OOXML types while still matching all standard XML content types.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1246,8 +1246,14 @@ func (c *Collector) handleOnXML(resp *Response) error {
 		return nil
 	}
 	contentType := strings.ToLower(resp.Headers.Get("Content-Type"))
+	// Parse the media type without parameters (e.g. charset)
+	mediatype, _, _ := strings.Cut(contentType, ";")
+	mediatype = strings.TrimSpace(mediatype)
 	isXMLFile := strings.HasSuffix(strings.ToLower(resp.Request.URL.Path), ".xml") || strings.HasSuffix(strings.ToLower(resp.Request.URL.Path), ".xml.gz")
-	if !strings.Contains(contentType, "html") && (!strings.Contains(contentType, "xml") && !isXMLFile) {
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Thanks!

PR #872, opened 2026-05-15
 Make (*LimitRule).Init idempotent — safe to share *LimitRule across Collectors

Shinku-Chen:
## Summary
  
  `(*LimitRule).Init()` rebuilds `r.waitChan` and reassigns
  `r.compiledRegexp` / `r.compiledGlob` on every call. When the same
  `*LimitRule` is passed to more than one `Collector` via `c.Limit(rule)`
  (a natural pattern when several Collectors share the same rate-limit
  policy), the second `c.Limit(rule)` call causes two distinct problems
  against the first Collector that is already running requests:

  1. **Deadlock.** The newly-allocated `waitChan` orphans the slots A's
     in-flight fetch goroutines have already acquired on the old channel.
     Their `defer func(r *LimitRule) { ... <-r.waitChan }(r)` resolves the
     field at defer-execution time, so they receive from the new (empty)
     channel and block forever. A's `Wait()` never returns and its workers
     permanently hold parallelism slots.

  2. **Data race.** `compiledRegexp` / `compiledGlob` are written by the
     second `Init()` while the first Collector's workers concurrently read
     them inside `LimitRule.Match()` (called from `httpBackend.GetMatchingRule`
     on every request). `go test -race` flags this clearly.

  ## Fix

  Make `Init()` idempotent. `waitChan` is the last field set on success, so
  it doubles as an "already-initialized" sentinel:

  ```go
  func (r *LimitRule) Init() error {
      if r.waitChan != nil {
          return nil
      }
      // ... compile patterns ...
      r.waitChan = make(chan bool, max(r.Parallelism, 1))
      return nil
  }

  Once a *LimitRule has been successfully initialized, every subsequent
  call is a no-op. A single Collector still allocates the channel exactly
  once; the public API is unchanged.

  Test

  Adds TestSharedLimitRule_AInitThenBInit_ANotStuck to colly_test.go:

  1. Collector A calls Limit(rule) and dispatches enough requests to
  occupy every slot in rule.waitChan (handlers block on a release
  channel).
  2. Collector B calls Limit(rule) — the trigger that re-enters Init()
  on the shared rule.
  3. The handlers are released and the test asserts that A.Wait() returns
  within 10s.

  Without the fix the test deadlocks at step 3 and fails on the 10s timeout
  under plain go test, and additionally trips the race detector under
  go test -race on the compiledRegexp / compiledGlob writes. With the
  fix it passes both modes in well under a second.

  Notes / risk

  - No public API change.
  - Single-Collector code is unaffected (Init still runs the full body
  exactly once).
  - A user who mutates rule.DomainGlob / rule.DomainRegexp after the
  first successful Init() and expects a later c.Limit(rule) to pick up
  the change would no longer see that effect. This was already
  ill-defined under the old behavior (the new pattern would not apply to
  an existing Collector either), and seems acceptable for an internal
  initialization hook.

### DIFF
--- a/http_backend.go
+++ b/http_backend.go
@@ -63,11 +63,25 @@ type LimitRule struct {
 	waitChan       chan bool
 	compiledRegexp *regexp.Regexp
 	compiledGlob   glob.Glob
+	lock           sync.Mutex
 }
 
-// Init initializes the private members of LimitRule
+// Init initializes the private members of LimitRule.
+//
+// Init is idempotent: once a LimitRule has been successfully initialized,
+// subsequent calls are no-ops. This makes it safe to share a single
+// *LimitRule across multiple Collectors via Collector.Limit — without this,
+// the second Collector's Limit() call would (a) replace waitChan and
+// orphan the first Collector's already-acquired slots, deadlocking its
+// in-flight requests' defer <-r.waitChan, and (b) race with concurrent
+// Match() reads of compiledRegexp / compiledGlob.
 func (r *LimitRule) Init() error {
-	r.waitChan = make(chan bool, max(r.Parallelism, 1))
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	if r.waitChan != nil {
+		return nil
+	}
 	hasPattern := false
 	if r.DomainRegexp != "" {
 		c, err := regexp.Compile(r.DomainRegexp)
@@ -88,7 +102,22 @@ func (r *LimitRule) Init() error {
 	if !hasPattern {
 		return ErrNoPattern
 	}
+	r.waitChan = make(chan bool, max(r.Parallelism, 1))
 	return nil
+}
+
+// Clone returns a shallow copy of the LimitRule containing only its
+// exported fields. Unexported state (waitChan, compiled patterns, lock)
+// is left zero so the returned rule is independent of the original and
+// must be re-initialized via Init before use.
+func (r *LimitRule) Clone() *LimitRule {
+	return &LimitRule{
+		DomainRegexp: r.DomainRegexp,
+		DomainGlob:   r.DomainGlob,
+		Delay:        r.Delay,
+		RandomDelay:  r.RandomDelay,
+		Parallelism:  r.Parallelism,
+	}
 }
 
 func (h *httpBackend) Init(jar http.CookieJar) {
--- a/colly_test.go
+++ b/colly_test.go
@@ -28,6 +28,8 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


Shinku-Chen:
LimitRule多次初始化导致任务死锁的问题解决

  var LimitRule = &colly.LimitRule{DomainGlob: "*", Parallelism: 50}
  全局只有一个指针,然后所有 collector 都 c.Colly.Limit(config.LimitRule)。
  
  死锁过程

  1. Collector A 初始化 → LimitRule.Init() → 创建 waitChan_A(size 50)
  2. Collector A 跑了一些请求,有 N 个进了 buffer (waitChan_A)
  3. Collector B 初始化 → LimitRule.Init() 又跑一遍 → r.waitChan 被换成全新的空 waitChan_B
  4. Collector A 那些已经在飞的请求,defer 走到 <-r.waitChan 时,读的是当前 r.waitChan 字段(指针上的)= waitChan_B,空的 → 永久阻塞
  5. 后续 Collector C/D/... 再 Init 一遍,把 waitChan_B 又换成 waitChan_C ... 雪球越滚越大
  6. 最后 50 个槽位的 channel 缓冲也填满后,新请求的入口 r.waitChan <- true 也阻塞 → 6500 goroutine 的来源


因此这里判定一下waitChan是否被创建,创建就不在重复创建覆盖

asciimoo (approved):
LGTM. @WGH- please check it and merge if you agree.

http_backend.go
WGH-, line 1: I don't like this approach for the following reasons.

First, this doesn't solve concurrent access. Calling `(*Collector).Limit` concurrently (e.g. multiple goroutines having each own collector instance) results in a data race. The following test fails under the race detector (`go test -race`):

```go
func TestSharedLimitRuleRace(t *testing.T) {
	rule := &LimitRule{DomainGlob: "*", Parallelism: 2}
	var wg sync.WaitGroup
	for range 2 {
		wg.Go(func() {
			a := NewCollector()
			a.Limit(rule)
		})
	}
	wg.Wait()
}
```

This can be easily solved by using `sync.Once`, but

My second point is, I think `LimitRule` shared between different collectors should not imply shared state. After all, it's called a rule (i.e. plain object simply describing the rules), not a limiter/tracker (something that would imply inner state, timers, counter, etc).

With that said, I'd prefer  `(*Collector).Limit` making a shallow copy of `LimitRule` object.

Shinku-Chen: I've updated the code commit and added **sync.Mutex** to fix the concurrency issue.

1. The data race problem has been resolved with sync.Mutex.
2. Regarding shared state:
I believe different Collectors **do need to share the same state**. The LimitRule is designed to restrict concurrency for a specific target. For example, if Website A only allows 1 concurrent request, but I have 5 Collectors targeting Website A, those 5 Collectors must share the same LimitRule to ensure only one request runs at a time.
3. I have also added a Clone function to perform a shallow copy. It is **the user's responsibility** to decide when to share the state between LimitRules and when to use independent state based on their actual needs.

Shinku-Chen: Additionally, my different Collectors use the same proxy IP, which also has its own concurrency limit. Therefore, we need to control the total concurrent requests across all 5 Collectors, instead of limiting each Collector individually — which would result in 5 concurrent requests running at the same time.

After comprehensive consideration, I configure the Limit rule here to restrict the usage of proxy IPs, instead of imposing restrictions based on individual websites.

WGH-: Why do you use different collectors in your scenario?

Shinku-Chen: When crawling data, I first need to locate the target user via the search list, then obtain articles published by the user, and finally extract article content. Therefore, four collections need to be created, all targeting the same website. Speed limits shall be configured on a per-website basis.

WGH-: I gave it another thought. Okay, that's fine.

After all, sharing `LimitRule` across different collectors _almost_ worked as is, provided you somehow avoided the data race. Changing how shared `LimitRule`s work would be a potentially breaking change.

WGH- (approved):

http_backend.go
Shinku-Chen: thanks

PR #873, opened 2026-05-17
drop deprecated rand.Seed call in httpBackend.Init

### LINKED ISSUE #820: Rand.Seed is deprecated and global rand functions are auto seeded since go 1.20
Rand.Seed is used in the httpbackend Init method.

[https://go.dev/doc/go1.20#library](https://go.dev/doc/go1.20#library
)
[https://github.com/golang/go/issues/54880](https://github.com/golang/go/issues/54880
)

c-tonneslan:
Closes #820.

\`math/rand\` self-seeds with a random value since Go 1.20 and the deprecation note on \`rand.Seed\` is pretty explicit. go.mod already requires go 1.24, so the call in \`httpBackend.Init\` is dead weight and gopls flags it as deprecated.

### DIFF
--- a/http_backend.go
+++ b/http_backend.go
@@ -92,7 +92,6 @@ func (r *LimitRule) Init() error {
 }
 
 func (h *httpBackend) Init(jar http.CookieJar) {
-	rand.Seed(time.Now().UnixNano())
 	h.Client = &http.Client{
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Thanks!

Issue #875, opened 2026-05-29
Cloned Collectors inherit checkRedirectFunc (so it uses the parent's AllowURLRevisit)


gabrielbergoc:
Hey Guys, awesome package, I've been using for a while now.

One of my scrapers broke overnight, with the error "already visited", but I had already encountered this on development, and setting it's `AllowURLRevisit` to `true` had solved it.
When debugging, I noticed the `Collector` in `checkRedirectFunc` had it set to `false`, and I discovered that it was the parent `Collector`, from whom I had cloned the one that revisits the same url multiple times.
I took a look at the source code, and the `Clone` function doesn't update the `HttpClient`'s `CheckRedirect` function, so that seems to be the issue.
The final nail in the coffin was setting the parent's `AllowURLRevisit` to true, and seeing the error vanishing again.

Here, the `Init` function initializing the `c.backend.Client.CheckRedirect` function:
<img width="699" height="655" alt="Image" src="https://github.com/user-attachments/assets/4547de36-ef1d-41be-80bf-1f1c2cc07c22" />

And the `Clone` function doesn't update it:
<img width="1072" height="930" alt="Image" src="https://github.com/user-attachments/assets/cf8b584f-0b00-401f-843b-84c60b89f575" />

Why it was working before, and why it stopped, I don't know...

PR #885, opened 2026-06-04
fix(collector): give Clone an independent http.Client and lock

Shinku-Chen:
https://github.com/gocolly/colly/issues/875

Clone() shallow-copied the parent's *httpBackend, so the clone shared
the parent's *http.Client. That client's CheckRedirect handler is set
in Init() as a closure over the *parent* Collector, so the clone's
CheckRedirect kept evaluating the parent's AllowURLRevisit and filter
state. Toggling AllowURLRevisit on the clone therefore had no effect
on redirects: a clone with AllowURLRevisit = true still failed with
"already visited" after the first redirect to a URL the parent had
seen. Cloning also reused the parent's *sync.RWMutex, letting clone
and parent contend for the callback-list lock.

Give each clone its own *httpBackend and *http.Client whose
CheckRedirect closure binds to the clone, plus its own *sync.RWMutex.
Transport, cookie jar and Timeout are still shared from the parent
so connection pooling and session cookies continue to flow across
collectors, and LimitRules are kept (the rules themselves are safe to
share because LimitRule.Init is idempotent — the slice header is
copied so per-collector Limit() calls do not race on append). Update
the Clone godoc to spell out what is shared vs independent.

Add TestCloneAllowURLRevisitIndependent: a clone with
AllowURLRevisit = true visits the same redirect endpoint twice. The
test fails with "already visited" on the previous implementation and
passes with the fix; it also asserts the parent's AllowURLRevisit
field is not mutated.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1417,10 +1417,13 @@ func (c *Collector) Cookies(URL string) []*http.Cookie {
 }
 
 // Clone creates an exact copy of a Collector without callbacks.
-// HTTP backend, robots.txt cache and cookie jar are shared
-// between collectors.
+// The visited-URL store, robots.txt cache, cookie jar and HTTP transport
+// are shared between collectors so cookies and the connection pool flow
+// across them. The *http.Client and the callback-protecting lock are
+// independent so per-collector settings like AllowURLRevisit take effect
+// on the clone's redirects without leaking back into the parent.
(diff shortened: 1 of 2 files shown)


Shinku-Chen:
same as
https://github.com/gocolly/colly/pull/882
https://github.com/gocolly/colly/pull/877

PR #890, opened 2026-06-15
Preserve Host when unmarshaling a request

Noethix55555:
`Request.Marshal` serializes the `Host` field (`serializableRequest.Host`), but `Collector.UnmarshalRequest` never copies it back into the returned `Request`. Any request with a custom virtual host loses it on a Marshal/Unmarshal round-trip, which is what persistent queues do, so the request is then sent to the wrong vhost.

Fix copies `req.Host` into the rebuilt `Request`, matching the other serialized fields.

To reproduce: build a `Request` with `Host` set, `Marshal()` it, `UnmarshalRequest()` it, and observe `Host` is empty. Added `TestRequestMarshalRoundtripHost` covering this; it fails without the change and passes with it. Full root package tests and `go vet` pass.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -628,6 +628,7 @@ func (c *Collector) UnmarshalRequest(r []byte) (*Request, error) {
 		Ctx:       ctx,
 		ID:        c.requestCount.Add(1),
 		Headers:   &req.Headers,
+		Host:      req.Host,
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
Thanks!

PR #895, opened 2026-06-17
do not treat 2xx responses as errors

Noethix55555:
`handleOnError` uses `203` as the success/error boundary:

```go
if err == nil && (c.ParseHTTPErrorResponse || response.StatusCode < 203) {
    return nil
}
if err == nil && response.StatusCode >= 203 {
    err = errors.New(http.StatusText(response.StatusCode))
}
```

So any 2xx status from 203 upward (e.g. `204 No Content`, `206 Partial Content`) is treated as an error: `OnError` fires and `OnResponse` does not. 203 is itself a success code. The boundary should be 300 so the whole 2xx range is success.

Added `TestCollectorVisitNoContent` (a 204 response calls `OnResponse`, not `OnError`); it fails on the current code and passes with the change. Full package tests pass.

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1317,10 +1317,10 @@ func (c *Collector) handleOnXML(resp *Response) error {
 }
 
 func (c *Collector) handleOnError(response *Response, err error, request *Request, ctx *Context) error {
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
You are right. Thanks.

PR #925, opened 2026-08-14
fix: escape multipart field names in PostMultipart to prevent header injection

### LINKED ISSUE #924: 【BUG】PostMultipart does not escape multipart field names
## Summary

`PostMultipart` builds multipart request bodies manually and writes map keys directly into the `Content-Disposition` header without quoting or escaping them.

If a field name contains newline characters, the generated multipart body can contain unintended header lines.

## Affected Code

In `colly.go`, `createMultipartReader()` writes the field name directly into the multipart part header:

```go
for contentType, content := range data {
	buffer.WriteString(dashBoundary + "\n")
	buffer.WriteString("Content-Disposition: form-data; name=" + contentType + "\n")
	buffer.WriteString(fmt.Sprintf("Content-Length: %d \n\n", len(content)))
	buffer.Write(content)
	buffer.WriteString("\n")
}
```

## Safe Validation Evidence

Tested locally against commit `20d31482af5f754832a753f88517f3dfa61d921f`.

```text
=== RUN   TestReproH3_MultipartCRLFInjection
    repro_high_test.go:65: H3 = TRUE POSITIVE: CRLF in data key injected X-Injected header into multipart (CWE-93)
--- PASS: TestReproH3_MultipartCRLFInjection (0.00s)
```

The current `master` branch still contains the same code path in `colly.go`.

## Expected Behavior

Multipart field names should be safely encoded according to multipart form-data rules.

## Actual Behavior

Field names are concatenated directly into the multipart header.

## Impact

Unexpected newline characters in field names can produce malformed multipart bodies and unintended part headers.

## Suggested Fix

Use Go's `mime/multipart.Writer` to construct multipart bodies, or quote and escape field names before writing them into `Content-Disposition`.

haoku123:
## Description

Fixes #924: `PostMultipart` writes map keys directly into the `Content-Disposition` header without quoting or escaping. A field name containing CR/LF could inject arbitrary header lines into the multipart body.

## Changes

- `colly.go`: new `escapeMultipartFieldName` helper implementing RFC 7578 §4.2 escaping (quote the value, backslash-escape `\\` and `"`, percent-encode CR/LF as `%0D`/`%0A`), applied in `createMultipartReader`.
- `colly_test.go`: `TestPostMultipartEscapesFieldNames` parses the generated body with `mime/multipart` and asserts:
  - normal / quote / backslash names round-trip correctly (Go's parser unescapes quoted-strings),
  - CR/LF in a name stays percent-encoded (`new%0D%0Aline`),
  - the raw body contains no unescaped newline in a field name.

## Verification

- `go test -run TestPostMultipartEscapesFieldNames`: PASS
- `go test ./...`: all packages PASS
- Control experiment: reverting the escaping makes the new test fail with `malformed MIME header: missing colon` (the header injection the issue describes)
- `go vet .`: clean

### DIFF
--- a/colly.go
+++ b/colly.go
@@ -1580,7 +1580,7 @@ func createMultipartReader(boundary string, data map[string][]byte) io.Reader {
 	buffer.WriteString("Content-type: multipart/form-data; boundary=" + boundary + "\n\n")
 	for contentType, content := range data {
 		buffer.WriteString(dashBoundary + "\n")
-		buffer.WriteString("Content-Disposition: form-data; name=" + contentType + "\n")
+		buffer.WriteString("Content-Disposition: form-data; name=\"" + escapeMultipartFieldName(contentType) + "\"\n")
 		buffer.WriteString(fmt.Sprintf("Content-Length: %d \n\n", len(content)))
 		buffer.Write(content)
 		buffer.WriteString("\n")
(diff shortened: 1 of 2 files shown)


asciimoo:
Good fix, thanks.

Minor note: In escapeMultipartFieldName (https://github.com/gocolly/colly/blob/ad96a72fb495949a05c313e29c04acbef9a5cfd8/colly.go#L1597-L1612), ranging over the string decodes UTF-8 invalid bytes are silently replaced with `U+FFFD`, changing field names that the previous implementation preserved.

Perhaps it would be better to use `strings.NewReplacer`, something like:
```go
var multipartFieldNameEscaper = strings.NewReplacer(
      "\\", "\\\\",
      `"`, `\"`,
      "\r", "%0D",
      "\n", "%0A",
)
```

What do you think?

haoku123:
Thanks for the review! Switched to `strings.NewReplacer` — it operates on the raw byte sequence and preserves invalid UTF-8 bytes. Added `TestEscapeMultipartFieldNamePreservesInvalidUTF8` to cover this.

asciimoo (approved):
LGTM. Thanks.

asciimoo:
Seems like the tests are failing.

haoku123:
Fixed the gofmt alignment in `colly_test.go` — the map keys weren't aligned to the longest key `new\r\nline`. CI should pass now.

PR #926, opened 2026-08-14
fix: apply MaxBodySize to the decompressed body, not the compressed stream

### LINKED ISSUE #923: 【BUG】MaxBodySize is applied before gzip decompression
## Summary

`MaxBodySize` is applied to the compressed response stream before Colly performs gzip decompression. As a result, a small compressed response can expand into a much larger response body after the limit has already been applied.

## Affected Code

In `http_backend.go`, the response body is wrapped with `io.LimitReader` before gzip detection and decompression:

```go
if bodySize > 0 {
	bodyReader = io.LimitReader(bodyReader, int64(bodySize))
}
...
bodyReader, err = gzip.NewReader(bufReader)
...
body, err := io.ReadAll(bodyReader)
```

This means the limit applies to compressed bytes, while `io.ReadAll` reads the decompressed stream.

## Safe Validation Evidence

Tested locally against commit `20d31482af5f754832a753f88517f3dfa61d921f`.

```text
=== RUN   TestReproH1_GzipBombUnbounded
    repro_high_test.go:27: H1 = TRUE POSITIVE: 2072-byte compressed -> 2097152 bytes decompressed via io.ReadAll after LimitReader (CWE-409)
--- PASS: TestReproH1_GzipBombUnbounded (0.01s)
```

The current `master` branch still contains the same code path in `http_backend.go`.

## Expected Behavior

`MaxBodySize` should limit the final response body size exposed to Colly callbacks, including after decompression.

## Actual Behavior

A gzip response can stay under `MaxBodySize` while compressed, then expand beyond that limit during decompression.

## Impact

Very small compressed responses can produce much larger in-memory response bodies than the configured `MaxBodySize` suggests.

## Suggested Fix

Apply a size limit to the decompressed stream as well, or enforce the limit while reading the final body.

haoku123:
## Description

Fixes #923: `MaxBodySize` is applied to the compressed response stream before gzip decompression, so a small gzip payload can expand far beyond the limit (CWE-409 decompression bomb).

## Root Cause

In `http_backend.go`, `io.LimitReader` wraps `res.Body` before gzip detection. When colly's own gzip branch runs (e.g. `.xml.gz` responses without `Content-Encoding`, or a client with compression disabled), `io.ReadAll` consumes the **decompressed** stream with no size limit.

Reproduced locally: 2 KiB compressed → 2 MiB decompressed body passed to callbacks with `MaxBodySize(1MiB)`, matching the issue's repro (2072 → 2097152 bytes).

## Changes

Keep the existing limit on the raw stream (first line of defense against malformed compression layers) and add a second `LimitReader` **after** the gzip branch, so the final body never exceeds `MaxBodySize`:

```go
if bodySize > 0 {
    bodyReader = io.LimitReader(bodyReader, int64(bodySize))
}
body, err := io.ReadAll(bodyReader)
```

## Verification

- New test `TestMaxBodySizeAfterDecompression` serves gzip data on a `.xml.gz` path (no `Content-Encoding` header, so Go's Transport does not auto-decompress and colly's gzip branch runs):
  - without the fix: `response body = 2097152 bytes, exceeds MaxBodySize 1048576` (FAIL)
  - with the fix: PASS, body truncated to the limit
- `go test ./...`: all packages PASS
- `go vet .`: clean

### DIFF
--- a/colly_test.go
+++ b/colly_test.go
@@ -2352,3 +2352,61 @@ func TestEscapeMultipartFieldNamePreservesInvalidUTF8(t *testing.T) {
 		t.Errorf("escapeMultipartFieldName replaced invalid UTF-8 with U+FFFD: %q", escaped)
 	}
 }
+func TestMaxBodySizeAfterDecompression(t *testing.T) {
+	// 2 MiB of data compresses to a few KiB, well under the wire limit.
+	big := bytes.Repeat([]byte("a"), 2*1024*1024)
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ww := gzip.NewWriter(w)
+		defer ww.Close()
+		ww.Write(big)
+	}))
+	defer ts.Close()
+
+	const limit = 1024 * 1024 // 1 MiB
+	c := NewCollector(MaxBodySize(limit))
+	var bodyLen int
+	c.OnResponse(func(r *Response) {
+		bodyLen = len(r.Body)
+	})
+	err := c.Visit(ts.URL + "/bomb.xml.gz")
+	if err != nil {
+		t.Fatalf("Visit failed: %v", err)
+	}
+	if bodyLen > limit {
+		t.Errorf("response body = %d bytes, exceeds MaxBodySize %d (decompression bomb not limited)", bodyLen, limit)
+	}
+	if bodyLen == 0 {
+		t.Errorf("response body is empty, expected up to %d bytes", limit)
+	}
+}
+
+// TestMaxBodySizeAllowsCompressedResponseWithinLimit verifies that a
+// compressed response whose decompressed body is under MaxBodySize is not
+// truncated by capping the compressed stream. gzip has ~18 bytes of
+// header/footer overhead, so an 8-byte plain body compresses to more bytes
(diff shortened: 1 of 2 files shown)


asciimoo:
This PR includes all of PR #925.

Also, the original `LimitReader` remains before decompression (https://github.com/gocolly/colly/blob/83c334fbe895da0eb94488531bb00da751dd58ad/http_backend.go#L234-L268), while
another is added afterward. Consequently, legitimate compressed responses can fail even when their decompressed body is below the configured limit.

haoku123:
Thanks for catching both issues!

1. **Rebased off master** — this PR no longer includes the #925 changes (my branch had been created off the multipart branch by mistake).
2. **Removed the pre-decompression LimitReader** — MaxBodySize is now applied only *after* gzip decompression, so it caps the final body without truncating legitimate compressed responses whose decompressed body is under the limit.

Added `TestMaxBodySizeAllowsCompressedResponseWithinLimit` to cover the regression: an 8-byte body with a 10-byte limit (gzip overhead pushes the wire size above the limit) decompresses correctly. Without the fix this fails with `unexpected EOF`.

haoku123:
Rebased onto latest master after #925 landed — the multipart tests are no longer duplicated in this branch, and only the MaxBodySize changes remain. Local full test suite passes.

PR #927, opened 2026-09-02
[fix] update to gobwas/glob 1.0

IceflowRE:


cristaloleg (approved):
This indeed fixes the issue. Thanks!

